PR #218, opened 2014-10-08
Use a windshaft-specific name for the test user

strk:
Avoids problems with running cartodb-sql-api tests after
windshaft-cartodb tests (as the test database is not cleaned up)


strk:
Thanks :)

PR #246, opened 2015-01-23
Drop signed maps

### LINKED ISSUE #238: Remove Signed Maps
Signed maps is a layer on top of template maps to handle authorisation over them. However the implementation is adding complexity to the problem it is solving and it is prone to race condition errors.

Let's remove it and keep it simple by checking the authorisation in the mapconfig object itself.

rochoa:
- Splits template maps endpoint into its own controller
 - Removes TemplateMaps dependency on SignedMaps
    - Token validation is done against the template
    - Template is always extended with default values for auth and placeholders
    - MapConfig is extended, in order to validate auth_tokens, with template info:
       - template name
       - template auth
    - No more locks to create, update or delete templates
       - Trusting in redis' hash semantics
       - Some tradeoffs:
           * A client having more templates than allowed by a race condition between limit (HLEN) check and creation (HSET)
           * Updating a template could happen while deleting it, resulting in a new template
           * Templates already instantiated will be accessible through their layergroup so it is possible to continue requesting tiles/grids/etc.
    - Authorisation is now handled by template maps
- Template instantiation returns new instances with default values if they are missing

This closes #238


rochoa:
cc @javisantana @zenitraM could you review this one?

zenitraM:
:+1:

PR #248, opened 2015-01-26
Basic layergroup validation on named map creation/update

rochoa:


lib/cartodb/template_maps.js
```diff
@@ -128,6 +133,18 @@ o._checkInvalidTemplate = function(template) {
     return false;
 };
 
+function isInvalidLayergroup(layergroup) {
+    if (!layergroup) {
+        return new Error('Missing layergroup');
+    }
+
+    if (!_.isArray(layergroup.layers)) {
```
javisantana: shouldn't we check that the layergroup contains at least 1 layer?

rochoa:
Now validation checks is an array, it's not an empty array and each layer has at least `options`, what do you think now? :smiley:

PR #249, opened 2015-01-26
Add named maps surrogate keys and call invalidation on template modification/deletion

rochoa:
- Extends TemplateMaps backend with EventEmitter
    - Emits for create, update and delete templates
 - VarnishHttpCacheBackend will invalidate a varnish instance via HTTP PURGE method
    - In the future there could be more backends, for instance to invalidate a CDN.
 - NamedMapsEntry has the responibility to generate a cache key for a named map
    - It probably should receive a template/named map instead of owner and template name
 - SurrogateKeysCache is resposible to tag responses with a header
    - It also is responsible for invalidations given an Invalidation Backend
    - In the future it could have several backends so it can invalidates different caches
- SurrogateKeysCache is subscribed to TemplateMaps events to do the invalidations

### DIFF
--- a/config/environments/development.js.example
+++ b/config/environments/development.js.example
@@ -150,6 +150,7 @@ var config = {
     ,varnish: {
         host: 'localhost',
         port: 6082,
+        http_port: 6081,
         secret: 'xxx',
         ttl: 86400,
         layergroupTtl: 86400 // the max-age for cache-control header in layergroup responses
--- a/config/environments/development.js.example
+++ b/config/environments/development.js.example
@@ -149,9 +149,9 @@ var config = {
     }
     ,varnish: {
         host: 'localhost',
-        port: 6082,
-        http_port: 6081,
-        purge_enabled: false,
+        port: 6082, // the por for the telnet interface where varnish is listening to
+        http_port: 6081, // the port for the HTTP interface where varnish is listening to
+        purge_enabled: false, // whether the purge/invalidation mechanism is enabled in varnish or not
         secret: 'xxx',
         ttl: 86400,
         layergroupTtl: 86400 // the max-age for cache-control header in layergroup responses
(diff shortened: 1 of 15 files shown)


rochoa:
please @zenitraM could you review this one?

lib/cartodb/cache/backend/varnish_http.js
```diff
@@ -0,0 +1,32 @@
...
+}
+
+module.exports = VarnishHttpCacheBackend;
+
+/**
+ * @param cacheObject should respond to `key() -> String` method
+ * @param {Function} callback
+ */
+VarnishHttpCacheBackend.prototype.invalidate = function(cacheObject, callback) {
+    request(
```
javisantana: why don't we follow the same thing fastly does ? https://docs.fastly.com/api/purge#purge_3

config/environments/development.js.example
```diff
@@ -150,6 +150,8 @@ var config = {
     ,varnish: {
         host: 'localhost',
         port: 6082,
+        http_port: 6081,
+        purge_enabled: false,
```
javisantana: some explanation would be good

lib/cartodb/cache/backend/varnish_http.js
zenitraM: It's not trivial to properly do the string concatenation that would be needed to make `key` purge `\bkey\b`.

We're doing the same way we have on the Rails apps but using a different endpoint which will match Surrogate-Key instead of X-Cache-Channel

javisantana: :+1:

config/environments/development.js.example
rochoa: I will take care ;-)

zenitraM:
:+1:

PR #307, opened 2015-06-04
Changes rules for names in templates

rochoa:
Now valid names can start with numbers and can contain dashes (-).


Kartones:
Apart from the failing specs, :+1:

rochoa:
Fixed!

PR #313, opened 2015-06-22
Update Map-API docs

andrewbt:
It was not clear to me what the difference was between a mapnik layer and a cartodb layer until I read the comments in the Mapconfig spec! Let's save some users the extra step :)


rochoa:
Thanks @andrewbt !!

PR #314, opened 2015-06-24
Use CDB_QueryTablesText instead of CDB_QueryTables

rafatower:
This avoids trouble with len(schema.table_name) > 63
See https://github.com/CartoDB/cartodb-postgresql/issues/86

This should be released after rolling out a new version of the cartodb-postgresql extension, with this PR merged: https://github.com/CartoDB/cartodb-postgresql/pull/88

@rochoa please review

### DIFF
--- a/lib/cartodb/api/query_tables_api.js
+++ b/lib/cartodb/api/query_tables_api.js
@@ -14,7 +14,7 @@ module.exports = QueryTablesApi;
 
 QueryTablesApi.prototype.getAffectedTablesInQuery = function (username, sql, callback) {
 
-    var query = 'SELECT CDB_QueryTables($windshaft$' + prepareSql(sql) + '$windshaft$)';
+    var query = 'SELECT CDB_QueryTablesText($windshaft$' + prepareSql(sql) + '$windshaft$)';
 
     this.pgQueryRunner.run(username, query, handleAffectedTablesInQueryRows, callback);
 };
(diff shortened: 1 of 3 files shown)


rochoa:
Is guaranteed to return always an array even if it's empty?

p.s.: I add a commit to point to the script with the new function, we must change once is merged and released.

rafatower:
> Is guaranteed to return always an array even if it's empty?
it is "by design" but don't know how to test the whole thing.

rafatower:
This is ready to be merged and used against cartodb postgres extension 0.8.0+

rafatower:
Bump!

rochoa:
I was thinking about this again and I see a problem with the "not always returning the table schema but when it is necessary". Why?

In Windshaft-cartodb we use the affected tables to add some headers so we can later invalidate resources based on those headers. Without the schema in the header when a table is modified and invalidation is triggered we have to invalidate by table_name instead of schema_name+table_name, that means that all resources associated with a table_name will get pruned from cache.

The biggest problem is when we add a header that is a message digest associated to the table name, because the message digest for a table_name with and without schema_name will be different. As an example:

User `foo` with a one layer visualization with `select * from table_name_1`, the affected tables without schema will be `['table_name_1']`, so the header will be something like: ```Cache-Key: `md5('table_name_1')`'``` but any other user having a table with name `table_name_1` will get its resources invalidated.

We don't apply the message digest algorithm to the table_name but a combination of table_name and database name but this still applies to users in the same database.

rochoa:
As a possible workaround we could use the db user for the message digest.

rafatower:
Sorry for the confusion:
- The implementation changed since the first time we've talked about it and finally it is adding the schema name as it used to: https://github.com/CartoDB/cartodb-postgresql/pull/88/files#diff-792ccbfa854d01427cd98a76e777a740R48

- The legacy function `CDB_QueryTables` is using the new function underneath, so if there's a problem, the problem is already there

- But no worries, it is not a problem :wink: 
```
# select cdb_querytables('select * from chicago_major_streets');
        cdb_querytables         
--------------------------------
 {public.chicago_major_streets}
(1 row)

cartodb_user_66461071-c2e5-43d4-9780-b0f12b098de6_db=# select cdb_version();
 cdb_version 
-------------
 0.8.1
```

rochoa:
Thank you for the clarification! :relieved:

PR #333, opened 2015-09-25
Remove outdated docs

rochoa:
I'm removing this old doc files, all that information is aggregated in the official doc file Map-API.md which gets published in http://docs.cartodb.com/cartodb-platform/maps-api.html

This old docs are usually confusing people about what documentation should be used, specially for named maps vs templates use case.

cc @ohasselblad @andrewbt what do you think?


ohasselblad:
I find it confusing, too! I support removing it.

andrewbt:
Makes sense to me. The Template-maps.md file all sounds important but if it's truly all duplicated on the Maps API page then it's probably OK to delete.

rochoa:
Thanks both!

@andrewbt It is duplicated, yes.

PR #351, opened 2015-12-31
Install instructions

pnorman:


pnorman:
I am not quite sure why the push CI build failed but the PR one passed. It looks like the push build ran a couple additional tests.

rochoa:
Ok from my side.

I re-triggered the tests, it seems like some unstable tests. We will have to review them if they start to be a common problem.

PR #365, opened 2016-01-20
Overviews work

jgoizueta:
This branch adds overviews metadata to the MapConfig so that Windshaft can use it for more efficient rendering. This modifications require the CDB_Overviews function to be defined in the cartodb
PostgreSQL extension.


jgoizueta:
@rochoa @rafatower please CR this / feel free to invite other reviewers

lib/cartodb/api/overviews_api.js
```diff
@@ -0,0 +1,58 @@
+var queue = require('queue-async');
+var QueryTablesApi = require('./query_tables_api');
+
+function OverviewsApi(pgQueryRunner) {
+   if (pgQueryRunner.pgQueryRunner !== undefined) {
```
rochoa: :see_no_evil: If this requires both collaborators I would declare the constructor with both of them.

It's hard to understand the naming and it's assuming QueryTablesApi has a PgQueryRunner instance.

lib/cartodb/api/overviews_api.js
```diff
@@ -0,0 +1,58 @@
...
+            callback(err);
+        } else {
+            var metadata = {};
+
+            var parallelism = 2;
+            var q = queue(parallelism);
+
+            tableNames.forEach(function(tableName) {
+                q.defer(function(done){
+                    var query = "SELECT * FROM CDB_Overviews('" + tableName + "');";
```
rochoa: Could we make CDB_Overviews to accept several tables? That way we could retrieve overviews for several layers at the same time, requiring just one query/connection to the database.

lib/cartodb/api/overviews_api.js
```diff
@@ -0,0 +1,58 @@
...
+                            done(new Error('could not get overviews metadata: ' + msg));
+                            return;
+                        }
+                        if ( rows.length > 0 ) {
+                          var table_metadata = {};
+                          for ( var i=0; i<rows.length; ++i ) {
+                            var row = rows[i];
+                            table_metadata[row.z] = { table: row.overview_table };
+                          }
+                          metadata[tableName] = table_metadata;
```
rochoa: Not a big fan of accessing metadata because it's available in this scope. Metadata aggregation should be done in `awaitAll`, where you will receive an array with all results for tasks in the queue.

lib/cartodb/controllers/base.js
```diff
@@ -249,7 +249,7 @@ function statusFromErrorMessage(errMsg) {
     else if (errMsg.match(/Postgis Plugin.*[\s|\n].*column.*does not exist/)) {
         statusCode = 400;
     }
-    else if ( -1 !== errMsg.indexOf('does not exist') ) {
+    else if ( -1 !== errMsg.indexOf('does not exist') && !errMsg.match(/function .* does not exist/ ) ) {
         if ( -1 !== errMsg.indexOf(' role ') ) {
```
rochoa: What about having the `function` case here?

lib/cartodb/models/mapconfig_overviews_adapter.js
```diff
@@ -0,0 +1,54 @@
...
+  var self = this;
+
+  if (!layers) {
+      return callback(null);
+  }
+
+  var augmentLayersQueue = queue(layers.length);
+
+  function augmentLayer(layer, done) {
+      if ( layer.type !== 'mapnik' && layer.type !== 'cartodb' ) {
```
rochoa: I know this is nothing new, but I'm starting to dislike the fact that you only receive the layers here, probably I makes more sense to receive the whole mapconfig object, because that has abstraction for checking layer types without having to know that `mapnik` and `cartodb` are the same type.

Don't worry about this, I will try to fix this.

lib/cartodb/models/mapconfig_overviews_adapter.js
```diff
@@ -0,0 +1,54 @@
...
+          return done(null, layer);
+      }
+      self.overviewsApi.getOverviewsMetadata(username, layer.options.sql, function(err, metadata){
+           if (err) {
+               done(err, layer);
+           } else {
+               if ( !_.isEmpty(metadata) ) {
+                   // layer.options.overviews = metadata;
+                   layer = _.extend({}, layer);
```
rochoa: I would prefer to not rely on modifying the layers here but return new array of layers so the client using the adapter can decide what to do with them.

test/acceptance/multilayer.js
```diff
@@ -617,7 +617,7 @@ describe(suiteName, function() {
           headers: {host: 'localhost', 'Content-Type': 'application/json' },
           data: JSON.stringify(layergroup)
       }, {}, function(res) {
-          assert.equal(res.statusCode, 404, res.statusCode + ": " + res.body);
+          assert.equal(res.statusCode, 400, res.statusCode + ": " + res.body);
```
rochoa: This makes sense, but... why did this change?

lib/cartodb/api/overviews_api.js
jgoizueta: Definitely, I would go for supporting multiple tables in one CDB_Overviews call

lib/cartodb/api/overviews_api.js
jgoizueta: Yes, this is ugly and probably brittle. I started with requiing a pgQueryRunner, then decided to accept a QueryTablesAPI too for convenience. It is the latter the only actualy used, so we can change to it.

jgoizueta: I propose to modify the OverviewsApi constructor to require a PgQueryRunner and a QueryTablesApi
(or to require a PgQueryRunner and build its own QueryTablesApi?)

In any case, then I think it would be better to add an overviewsApi parameter to MapController
and instantiate the OverviewsApi in servier.js

lib/cartodb/models/mapconfig_overviews_adapter.js
jgoizueta: That misleading comment (layer.options.overviews = ...) must be removed. A new layers array is actually being built here, copying existing layer attributes into new layer objects.

test/acceptance/multilayer.js
jgoizueta: The error cause is a non-existent SQL function (bogus). That used to be assigned a 404 because the error message included  "does not exist" and now we've changed that to recognise "function ... does not exist" and make that a 400.

lib/cartodb/api/overviews_api.js
jgoizueta: I'll review that

lib/cartodb/controllers/base.js
jgoizueta: Yes, I'll look into it

jgoizueta: Mmm I misunderstood your comment, you meant why did I put the `/function...` check just there? (It was just to quiet jshint: 81cb75f82)

rochoa: Oh, cyclomatic complexity. But ... isn't it clearer the other way around?

On Fri, Jan 22, 2016 at 11:09 AM, Javier Goizueta <notifications@github.com>
wrote:

> In lib/cartodb/controllers/base.js
> <https://github.com/CartoDB/Windshaft-cartodb/pull/365#discussion_r50520299>
> :
>
> > @@ -249,7 +249,7 @@ function statusFromErrorMessage(errMsg) {
> >      else if (errMsg.match(/Postgis Plugin.*[\s|\n].*column.*does not exist/)) {
> >          statusCode = 400;
> >      }
> > -    else if ( -1 !== errMsg.indexOf('does not exist') ) {
> > +    else if ( -1 !== errMsg.indexOf('does not exist') && !errMsg.match(/function .* does not exist/ ) ) {
> >          if ( -1 !== errMsg.indexOf(' role ') ) {
>
> Mmm I misunderstood your comment, you meant why did I put the /function...
> check just there? (It was just to quiet jshint: 81cb75f
> <https://github.com/CartoDB/Windshaft-cartodb/commit/81cb75f821807ef8fc7fc6bb6ac59a5626ff7de1>
> )
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/CartoDB/Windshaft-cartodb/pull/365/files#r50520299>.
>



-- 
Raul Ochoa
rochoa@cartodb.com

rafatower:
Looks good to me (though took me a while to understand it, especially the `mapconfig_overviews_adapter.js` async part). Anyway I'm just a "voyeur" here and have nothing to contribute.

jgoizueta:
please retest this

jgoizueta:
Windshaft version has to be updated to one supporting QueryRewriter. Since none is release yet, can use this:

```
npm install https://github.com/CartoDB/Windshaft/tarball/overviews-work —save
```

But I've found problems trying to shrinkwrap this:

```
npm ERR! Problems were encountered
npm ERR! Please correct and try again.
npm ERR! invalid: readable-stream@1.0.27-1 /home/ubuntu/www/node-windshaft/current/node_modules/windshaft/node_modules/grainstore/node_modules/millstone/node_modules/sqlite3/node_modules/node-pre-gyp/node_modules/tar-pack/node_modules/readable-stream
npm ERR! invalid: string_decoder@0.10.25-1 /home/ubuntu/www/node-windshaft/current/node_modules/windshaft/node_modules/grainstore/node_modules/millstone/node_modules/sqlite3/node_modules/node-pre-gyp/node_modules/tar-pack/node_modules/readable-stream/node_modules/string
```

So `package.json` and `npm-shrinkwrap.json` are not updated here and the CI tests fail.

lib/cartodb/controllers/map.js
```diff
@@ -203,7 +223,23 @@ MapController.prototype.instantiateTemplate = function(req, res, prepareParamsFn
             );
             mapConfigProvider.getMapConfig(this);
         },
-        function createLayergroup(err, mapConfig_, rendererParams/*, context*/) {
+        function addOverviewsInformation(err, requestMapConfig, rendererParams/*, context*/) {
```
rafatower: deja vu (also createLayergroup)

lib/cartodb/utils/overviews_query_rewriter.js
```diff
@@ -0,0 +1,197 @@
...
+// of the form detected by the overviews_supported_query function.
+OverviewsQueryRewriter.prototype.query = function(query, data) {
+    var overviews = this.overviews_metadata(data);
+    if ( !overviews || !this.is_supported_query(query)) {
+        return query;
+    }
+    var zoom_level_expression = this.options.zoom_level || '0';
+    return overviews_query(query, overviews, zoom_level_expression);
+};
+
```
rafatower: A matter of personal taste: to see the big picture I prefer the code top-down, from public functions down to helper functions. IMHO it makes it more readable.

jgoizueta: Good suggestion, I agree it helps

lib/cartodb/controllers/map.js
jgoizueta: Doing it all twice doesn't make for a more robust solution? :grin: That's something hard to catch with tests, thanks!

rafatower:
:+1:

PR #369, opened 2016-02-03
Geojson renderer

dgaubert:
We have to release Windshaft before and upgrade to the new version.

cc/ @rochoa


dgaubert:
We have to release Windshaft before and update to the new version.

rochoa:
OK!

dgaubert:
Released Windshaft 1.8.0. Can I merge?

test/acceptance/named_layers.js
```diff
@@ -405,7 +405,7 @@ describe('named_layers', function() {
                 var next = this;
                 assert.response(server,
                     {
-                        url: '/api/v1/map',
+                        url: '/api/v1/map?map_key=1234',
```
rochoa: If we had to change this is because something changed in layers so now they don't work without api keys. We need to review this before merging.

rochoa:
:ship: :it: !

PR #374, opened 2016-02-09
Use new CDB_QueryTablesUpdatedAt function

zenitraM:
The new CDB_QueryTablesUpdatedAt function returns schema_name, table_name, db_name, updated_at. 
This opens the possibility to have tables coming from different databases (as is the case of using Foreign Data Wrappers).

TODO:
- [x] Specs still run against the current functions on the branch "cdb" and will now fail on CI - those need to be changed those to install the `CDB_TableMetadata.sql` and `CDB_ForeignTables.sql` functions and then specs run just fine. _Done and now pointing to master_
- Now only the new function is used everywhere. We could add some logic to cache affected tables and then request updated_at times on demand, that should probably belong inside the QueryTables class to be shared with the SQL API.
- [x] X-Cache-Channel now appends multiple databases as: `db1:table1,table2;;db2:table1,table2,table3`. Changing the invalidation regex from `^dbname` to `(^|;;)dbname` should be enough to make invalidation work. **WIP: https://github.com/CartoDB/cartodb/pull/6559 **
- [x] Now all cache keys take into account the _really fully qualified name_: schema_name.table_name, even if schema_name should be hidden.
This is due to the new function not returning "pretty/regclass" names: we only have access to the stringified schema name and table name. This is fine, but we need to either either:
 - Change the function to return a single "regclass" table_name instead of separate schema_name and table_name (so we tag as `untitled_table` instead of `public.untitled_table`)
 - or tweak the invalidation triggers to always send the fully qualified name while invalidating (probably the more explicit and clean option). **Verified: we are already doing this**

cc @rochoa @rafatower


rafatower:
> or tweak the invalidation triggers to always send the fully qualified name while invalidating (probably the more explicit and clean option)

I prefer that option because beautified names from "regclass" depend a lot on the environment (`search_path` and checking the catalog to covert a local oid into the fully qualified name of a foreign table) but don't know how much work would that entail.

zenitraM:
> I prefer that option because beautified names from "regclass" depend a lot on the environment (search_path and checking the catalog to covert a local oid into the fully qualified name of a foreign table) but don't know how much work would that entail.

It turns out _we don't have to do anything_. QueryTablesText returns the fully qualified name already, in return X-Cache-Channel is already being sent using the fully qualified name.

The only thing that breaks the plan is escaping. Right now an schema with unauthorized characters or a reserved name (i.e 'zenitram-test') will have to be quoted and its QueryTablesText will return `"zenitram-test".tablename`.

However, CDB_Query_Tables_Updated_At will return schema_name as `fo-bar`, unquoted. (the same might happen with a table name with hyphens, even though the editor forbids using those).

To be consistent, is there any way for the `schema_name` and `table_name` columns to contain the quotes if the schema name / table name needs escaping?

pnorman:
[`quote_ident(string text)`](http://www.postgresql.org/docs/current/static/functions-string.html#FUNCTIONS-STRING-OTHER) will

> Return the given string suitably quoted to be used as an identifier in an SQL statement string. Quotes are added only if necessary (i.e., if the string contains non-identifier characters or would be case-folded). Embedded quotes are properly doubled.

zenitraM:
Moving to #383.

PR #383, opened 2016-02-22
Use new QueryTables library

zenitraM:
Uses the new library for affected table detection and cache table detection:
 https://github.com/CartoDB/node-cartodb-query-tables

Supersedes #374


zenitraM:
Oops, I messed up a git push - this is not really merged.

PR #386, opened 2016-02-23
Change error messages when getting overviews metadata fails

### LINKED ISSUE #384: Misleading error message for non-existent tables
If a layer contains a query that reference a non-existent table the error message is currently this:

```
could not get overviews metadata: relation \"xxxxx\" does not exist
```

Note that the main error statement says that overviews metadata wasn't obtained and the (most relevant) fact that the table does not exist appears as a detail.

The error message would be more clear to the user if it simply stated that the table does not exist.

The fact that overviews metadata must be fetched is a detail implementation and we should only show it in the error message if we cannot determine other cause for the problem.

jgoizueta:
Remove the detail that the error occurred trying to get overviews
metadata from the error message. This should be less confusing
to the user.

Fixes #384

### DIFF
--- a/lib/cartodb/api/overviews_metadata_api.js
+++ b/lib/cartodb/api/overviews_metadata_api.js
@@ -25,8 +25,7 @@ OverviewsMetadataApi.prototype.getOverviewsMetadata = function (username, sql, c
     var query = 'SELECT * FROM CDB_Overviews(CDB_QueryTablesText($windshaft$' + prepareSql(sql) + '$windshaft$))';
     this.pgQueryRunner.run(username, query, function handleOverviewsRows(err, rows) {
         if (err){
-            var msg = err.message ? err.message : err;
-            callback(new Error(msg));
+            callback(err);
             return;
         }
         var metadata = {};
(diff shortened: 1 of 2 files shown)


jgoizueta:
cc @javisantana @rochoa please review this

lib/cartodb/api/overviews_metadata_api.js
```diff
@@ -26,7 +26,7 @@ OverviewsMetadataApi.prototype.getOverviewsMetadata = function (username, sql, c
     this.pgQueryRunner.run(username, query, function handleOverviewsRows(err, rows) {
         if (err){
             var msg = err.message ? err.message : err;
-            callback(new Error('could not get overviews metadata: ' + msg));
```
rochoa: Shouldn't be enough to rethrow `err`?

javisantana:
I don't get this PR, how changing the text msg fixes the problem? what are we sending to the user when the sql contains tables that don't exist?

jgoizueta:
A 'relation does not exist' or similar error.

When the query for overview tables fails we were putting a "couldn't get overviews data" message in front of the error which actually caused the problem.
Now we just show that upstream message to the user (relation does not exist or whatever) without decoration.

lib/cartodb/api/overviews_metadata_api.js
jgoizueta: I'm not sure what you mean by rethrow here... we're passing the error through callbacks..., not using exceptions (?)

jgoizueta: Oh, sorry, now I see it, wraping it in Error...
let me think about it

jgoizueta: Yes, that made sense when we were taking the message and altering it, now it is probably better to pass along the `err`, I'll do it

PR #388, opened 2016-02-23
Use new querytables library

zenitraM:
I messed up #383 and now it shows as merged even if it is not :sweat_smile:  

Uses the new library for affected table detection and cache table detection:
https://github.com/CartoDB/node-cartodb-query-tables

Supersedes #374


rochoa:
I think we are in good shape to merge it in the current status :+1:

PR #396, opened 2016-03-07
Added config parameter to windshaft

dgaubert:
Added config parameter to windshaft in order to use postgres' function ST_RemoveRepeatedPoints (set false by the default)

cc @rochoa

### DIFF
--- a/lib/cartodb/server_options.js
+++ b/lib/cartodb/server_options.js
@@ -75,8 +75,7 @@ module.exports = {
                         idleTimeout: 3000,
                         reapInterval: 1000
                     },
-                    clipByBox2d: false,
-                    removeRepeatedPoints: false
+                    clipByBox2d: false
                 }
             }),
             torque: rendererConfig.torque,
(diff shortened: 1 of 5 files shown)


lib/cartodb/server_options.js
```diff
@@ -71,11 +71,12 @@ module.exports = {
...
                     dbPoolParams: {
-                          size: 16,
-                          idleTimeout: 3000,
-                          reapInterval: 1000
+                        size: 16,
+                        idleTimeout: 3000,
+                        reapInterval: 1000
                     },
                     clipByBox2d: false,
+                    removeRepeatedPoints: false
```
rochoa: Why do we need it here? Shouldn't be enough to rely on the default behaviour?

dgaubert: Actually we don't. But I wanted to be consistent and I did the same that I did with `clipByBox2d`. I'm going to remove it.

rochoa:
Superseded by #442.

Issue #405, opened 2016-03-29
Overviews queries don't support named maps wrapping


jgoizueta:
Overviews are applied currently only to queries of the form `SELECT ... FROM table`, but queries in named maps [are wrapped](https://github.com/CartoDB/cartodb/blob/132d4589c19cfa47826e20f617a74074b364b049/services/named-maps-api-wrapper/lib/named-maps-wrapper/named_map.rb#L285) as `SELECT * FROM (...) AS wrapped_query WHERE ...=1`

Issue #400, opened 2016-03-11
Fix overviews integration for named layers


dgaubert:
Overview integration is not working properly for named layers. Always receive [undefined layers](https://github.com/CartoDB/Windshaft-cartodb/blob/master/lib/cartodb/controllers/map.js#L224) from argument `requestMapConfig` due to is a `MapConfig` model instead a raw map config from requested body. It might integration for named layer should be placed in `name_map_provider` as we've already done for [turbo-cartocss](https://github.com/CartoDB/Windshaft-cartodb/pull/399) integration.

cc/ @jgoizueta & @rafatower

PR #406, opened 2016-03-31
[DO NOT MERGE] Overviews fixes

jgoizueta:
Fix overviews related issues #400 & #405


jgoizueta:
Should not mix multile issues in a single PR, will open separate branches/PRs

PR #407, opened 2016-04-01
Fix overviews integration for named layers

### LINKED ISSUE #400: Fix overviews integration for named layers

jgoizueta:
Fix #400

### DIFF
--- a/lib/cartodb/cache/named_map_provider_cache.js
+++ b/lib/cartodb/cache/named_map_provider_cache.js
@@ -7,12 +7,13 @@ var queue = require('queue-async');
 
 var LruCache = require("lru-cache");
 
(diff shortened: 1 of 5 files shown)


jgoizueta:
@dgaubert @rtorre please take a look

dgaubert:
:+1:

PR #409, opened 2016-04-01
Support overviews for named layer wrapped queries

### LINKED ISSUE #405: Overviews queries don't support named maps wrapping

jgoizueta:
Fixes #405

I've looked into adding aggregation-counter to overviews queries and flexibilizing the form of accepted queries, but at the moment I've opted for solving the current issue with minimal effort/changes.

### DIFF
--- a/lib/cartodb/utils/overviews_query_rewriter.js
+++ b/lib/cartodb/utils/overviews_query_rewriter.js
@@ -168,9 +168,16 @@ OverviewsQueryRewriter.prototype.query = function(query, data) {
 };
 
 OverviewsQueryRewriter.prototype.is_supported_query = function(sql) {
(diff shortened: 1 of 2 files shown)


jgoizueta:
@dgaubert @rafatower please CR this

rafatower:
:+1:

PR #421, opened 2016-04-19
Support unneeded schema names in overviews queries

### LINKED ISSUE #420: Overviews not applied if unnecessary schema name is included in queries
If a layer's query references a table including the schema name, and the schema is not required given the `schema_path` in use (i.e. if the table regclass casted to text doesn't include the schema) overview tables are not applied to the query.

This is a problem because, for organisation users, the editor adds the schema part (user name) to
table references when the query SQL is edited manually.

jgoizueta:
Fixes #420

Keep table schema of overviews base tables and use it
to support queries that use the schema name when not
strictly needed.

### DIFF
--- a/lib/cartodb/api/overviews_metadata_api.js
+++ b/lib/cartodb/api/overviews_metadata_api.js
@@ -1,5 +1,5 @@
 function OverviewsMetadataApi(pgQueryRunner) {
-   this.pgQueryRunner = pgQueryRunner;
+    this.pgQueryRunner = pgQueryRunner;
 }
 
 module.exports = OverviewsMetadataApi;
--- a/lib/cartodb/api/overviews_metadata_api.js
+++ b/lib/cartodb/api/overviews_metadata_api.js
@@ -22,7 +22,7 @@ function prepareSql(sql) {
 }
 
 OverviewsMetadataApi.prototype.getOverviewsMetadata = function (username, sql, callback) {
-    // FIXME: Currently using internal function parsed_table
+    // FIXME: Currently using internal function _cdb_schema_name
     //        CDB_Overviews should provide the schema information directly.
     var query = 'SELECT *, _cdb_schema_name(base_table)' +
                 '  FROM CDB_Overviews(CDB_QueryTablesText($windshaft$' + prepareSql(sql) + '$windshaft$))';
--- a/lib/cartodb/api/overviews_metadata_api.js
+++ b/lib/cartodb/api/overviews_metadata_api.js
@@ -31,18 +31,17 @@ OverviewsMetadataApi.prototype.getOverviewsMetadata = function (username, sql, c
             callback(err);
             return;
         }
-        var metadata = {};
-        rows.forEach(function(row) {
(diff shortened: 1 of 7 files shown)


jgoizueta:
@rochoa @rafatower can you please review this?

rafatower:
what you mean is that this makes the code independent of whether the schema is in the search path or not, right?

rafatower:
uhm, no, what you mean is that the schema can be provided even if not needed (when it is already in the search path). Correct me if I'm wrong.

:+1:

jgoizueta:
Yes, that's it: without this changes, if the schema is present when not needed (because it is in the search path and no other schema leading to a table with the same name is ahead of it in the path) overviews would not be used. Now overviews are applied in that case with or without having the explicit schema in the table name.

lib/cartodb/api/overviews_metadata_api.js
```diff
@@ -31,11 +34,13 @@ OverviewsMetadataApi.prototype.getOverviewsMetadata = function (username, sql, c
         var metadata = {};
         rows.forEach(function(row) {
```
dgaubert: Can we use `reduce` approach here?. For instance:

```
var metadata = rows.reduce(function (metadata, row) {
    metadata = ....
    ...
    return metadata;
}, {});
```

test/unit/cartodb/overviews_query_rewriter.js
```diff
@@ -202,6 +202,78 @@ describe('Overviews query rewriter', function() {
     done();
   });
 
+  it('uses schema name from overviews', function(done){
```
dgaubert: It seems a synchronous test, in that case, `done`callback is not needed.

dgaubert:
if you can fix two spaces indentation would be nice! ;)

lib/cartodb/api/overviews_metadata_api.js
jgoizueta: sure thing

jgoizueta:
Man there was a ton of bad indentation! :sweat:

So, after the next comment the overall diff will be useless...

dgaubert:
Ok, don't do it then! We will fix later then ;)

jgoizueta:
just did it!

lib/cartodb/api/overviews_metadata_api.js
```diff
@@ -22,22 +22,26 @@ function prepareSql(sql) {
...
         if (err){
             callback(err);
             return;
         }
-        var metadata = {};
-        rows.forEach(function(row) {
+        var metadata = rows.reduce(function(metadata, row){
             var table = row.base_table;
+            var schema = row._cdb_schema_name;
             var table_metadata = metadata[table];
```
dgaubert: I'm not sure what `table_metadata` does. I mean, it is only used to check `metadata[table]`, right?. I think we can remove it!

jgoizueta: Right, can be removed, I'm on it

dgaubert:
:+1:

PR #430, opened 2016-04-26
Add suport for WHERE clauses in overview queries

### LINKED ISSUE #428: Allow basic filtering in overviews queries
Currently when (vector) overviews exist for a table, they're used in the query passed to Mapnik only if the original query is of a very simple form not having a WHERE clause.

Allow the queries to have a simple (no joins, subselections, etc.) WHERE clause and apply it to the query passed to the tiler.

jgoizueta:
Fixes #428

### DIFF
--- a/lib/cartodb/utils/overviews_query_rewriter.js
+++ b/lib/cartodb/utils/overviews_query_rewriter.js
@@ -1,3 +1,4 @@
+
 var TableNameParser = require('./table_name_parser');
 
 function OverviewsQueryRewriter(options) {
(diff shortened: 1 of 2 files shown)


jgoizueta:
@rochoa can you review (or invite someone to) this? cc @rafatower

rochoa:
At some point we should start considering https://github.com/lfittl/pg_query as the alternative to this regex solution.

rochoa:
Understanding and maintaining this will get crazy otherwise.

:shipit:

jgoizueta:
I was thinking the same. Actually I did some testing with pg_query some time ago.

rochoa:
BTW, I'm missing the point of this PR. What do we want to achieve with it?

jgoizueta:
I did it with the intention of allowing filters (e.g. from the editor), but I'm focus now on having overviews work with widgets (so that filters applied by widgets apply to overviews visualization and so that overviews can by used by the dataviews used in widgets).

This approach is useless for those purposes, so I'm closing this now.

PR #435, opened 2016-05-03
link to named map with a torque layer block page

csobier:
For [Docs issue#822](https://github.com/CartoDB/docs/issues/822). Added clarification and link to example of a Torque layer in Named map, as per a reported Support inquiry.

@ztephm , does this help or do we need more information?

### DIFF
--- a/docs/named_maps.md
+++ b/docs/named_maps.md
@@ -451,6 +451,7 @@ callback({
 ```
 
 ## CartoDB.js for Named Maps
+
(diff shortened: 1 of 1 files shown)


ztephm:
thanks @csobier ! Link is helpful, but can we also add a section showing the torque code so that it's easier to compare to the non-torque structure on the same page?

csobier:
Makes sense, check out the updated. I added "Torque Layer as a Named Map" with the code, and used {placholders} where Carla's username was. The link to her blocks page is still in the note. IS there more code, or descriptions, needed?

ztephm:
Looks good, thanks!

PR #449, opened 2016-05-12
Use overviews for widgets (dataviews, filtered queries)

jgoizueta:
This adds support so that overviews are used for queries filtered by queries and then can also used in the dataviews.


lib/cartodb/utils/overviews_query_rewriter.js
```diff
@@ -1,5 +1,10 @@
+var _ = require('underscore');
 var TableNameParser = require('./table_name_parser');
 
+// this is meant as a hack for development (otherwise we'd be coupling
+// this to Camshaft impolementation)
+var queryBuilder = require('camshaft/lib/filter/query-builder');
```
jgoizueta: @rochoa: I've been using this for testing the code. (letting camshaft apply the filters to overview queries). What do you think we could do to put this in production? I don't feel comfortable with either replicating the filter-apply code from camshaft or introducing dependencies to the innards of Camshaft...

rochoa: IMHO

> duplication is far cheaper than the *wrong* abstraction

http://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction

jgoizueta: These are options I can think of:

* Current approach (using Camshaft's internal QueryBuilder): would require exposing this or a similar interface as a public API in Camshaft, which doesn't seem desirable.
* Duplicating the filter-application done in Camshaft: the problem with this is we're interpreting the filters data structure which is a Camshaft internal detail.
* Having a public interface in Camshaft (a Filter class?) that can take filters definitions and encapsulate in an object with methods to apply them to a query. We could have Node's getFilters() to return such Filter objects.

@rochoa: do you think we should go with the second as an interim solution? any chance of having the third?

rochoa: As I said I don't want to start abstracting things / exposing interfaces from camshaft. Once we commit to that we will have to live/support those abstractions, and I don't think we want to get there for now.

Have in mind that filters, as you are getting them from camshaft, are plain objects which are part of the analysis definition camshaft spec (Yes, I need to document this, sorry). And those objects are actually created at windshaft-cartodb:https://github.com/CartoDB/Windshaft-cartodb/blob/1c2354dc497819b7d5cea078534026ad5192a81b/lib/cartodb/models/analysis-mapconfig-adapter.js#L58-L70.

lib/cartodb/models/dataview/overviews/formula.js
```diff
@@ -0,0 +1,105 @@
+var _ = require('underscore');
+var BaseWidget = require('../base');
+var BaseDataview = require('../formula');
+
+var debug = require('debug')('windshaft:widget:formula:overviews');
```
ethervoid: This is production ready?

lib/cartodb/models/filter/camshaft/category.js
```diff
@@ -0,0 +1,79 @@
+'use strict';
+
+var debug = require('debug')('windshaft:filter:category');
```
ethervoid: Same here

lib/cartodb/models/dataview/overviews/formula.js
jgoizueta: mmm, I think no debug messages are left, but the module is required. I don't think it hurts but I'll remove it

lib/cartodb/backends/dataview.js
```diff
@@ -226,6 +268,8 @@ DataviewBackend.prototype.search = function (mapConfigProvider, user, params, ca
                 query = node.getQuery(applyFilters);
             }
 
+            // TODO: should handle overviews as getDataview ?
```
jgoizueta: I have yet to take a look at the search endpoint to see what it does. @rochoa maybe you could explain quickly its purpose?

rochoa: It allows to search by the defined column in the dataview. Probably it's easier if you check this: https://github.com/CartoDB/Windshaft-cartodb/blob/cfdac1bcb00e4143f1ab8aec4bd99fd962f41a32/lib/cartodb/models/dataview/aggregation.js#L217-L234

lib/cartodb/models/dataview/overviews/formula.js
rochoa: Don't worry about that. Keep them if you find them useful.

lib/cartodb/api/filter_stats_api.js
```diff
@@ -0,0 +1,63 @@
...
+var CamshaftFilter = require('../models/filter/camshaft');
+
+function FilterStatsApi(pgQueryRunner) {
+    this.pgQueryRunner = pgQueryRunner;
+}
+
+module.exports = FilterStatsApi;
+
+function getEstimatedRows(pgQueryRunner, username, query, callback) {
+    pgQueryRunner.run(username, "EXPLAIN "+query, function(err, result_rows) {
```
rochoa: What about using `EXPLAIN (FORMAT JSON) ...`?

jgoizueta: I'll try it (I didn't realize JSON was available, had only XML in mind which is used in the cartodb extension)

lib/cartodb/backends/dataview.js
```diff
@@ -109,11 +116,46 @@ DataviewBackend.prototype.getDataview = function (mapConfigProvider, user, param
                 query = node.getQuery(applyFilters);
             }
 
+            var sourceId =  dataviewDefinition.source.id; // node.id
+            var layer = _.find(
```
rochoa: `Array.prototype.find` does the same.

lib/cartodb/backends/dataview.js
```diff
@@ -109,11 +116,46 @@ DataviewBackend.prototype.getDataview = function (mapConfigProvider, user, param
...
+                mapConfig.obj().layers,
+                function(l){ return l.options.source && (l.options.source.id === sourceId); }
+            );
+            var queryRewriteData = layer && layer.options.query_rewrite_data;
+            if ( queryRewriteData ) {
+                if ( node.type === 'source' ) {
+                    var filters = node.filters; // TODO: node.getFilters() when available in camshaft
+                    var filters_disabler = _.keys(filters).reduce(
```
rochoa: `Object.keys().reduce`.

lib/cartodb/backends/dataview.js
jgoizueta: OK, take note.

lib/cartodb/utils/overviews_query_rewriter.js
```diff
@@ -148,34 +177,127 @@ function overviews_query(query, overviews, zoom_level_expression) {
...
+//             { 'table-name': {1: { table: 'overview-table-1' }, ... }, ... },
+//        zoom_level: ...,       // optional zoom level
+//        filters: ...,          // filters definition
+//        unfiltered_query: ..., // query without the filters
+//        bbox_filter: ...       // bounding-box filter
+//    }
+OverviewsQueryRewriter.prototype.query = function(query, data, options) {
+    options = options || {};
+
+    var overviews        = data && data.overviews;
```
rochoa: I would do a `data = data || {}` to avoid all these checks.

rochoa:
All my comments are _nice to have_ things, so go ahead.

I would also prefer camelCase over snake_case for functions.

lib/cartodb/api/filter_stats_api.js
```diff
@@ -0,0 +1,59 @@
...
+            rows = result_rows[0]['QUERY PLAN'][0].Plan['Plan Rows'];
+        }
+        return callback(null, rows);
+    });
+}
+
+FilterStatsApi.prototype.getFilterStats = function (username, unfiltered_query, filters, callback) {
+  var stats = {};
+  var self = this;
+  step(
```
rochoa: This could be done with a queue instead of step. Cosmetic change, so don't worry.

lib/cartodb/backends/dataview.js
jgoizueta: What should be our rule in these cases? always favor ES5 available methods over underscore?

rochoa: Yes :).

On Wed, May 18, 2016 at 2:52 PM, Javier Goizueta <notifications@github.com>
wrote:

> In lib/cartodb/backends/dataview.js
> <https://github.com/CartoDB/Windshaft-cartodb/pull/449#discussion_r63696279>
> :
>
> > @@ -109,11 +116,46 @@ DataviewBackend.prototype.getDataview = function (mapConfigProvider, user, param
> >                  query = node.getQuery(applyFilters);
> >              }
> >
> > +            var sourceId =  dataviewDefinition.source.id; // node.id
> > +            var layer = _.find(
> > +                mapConfig.obj().layers,
> > +                function(l){ return l.options.source && (l.options.source.id === sourceId); }
> > +            );
> > +            var queryRewriteData = layer && layer.options.query_rewrite_data;
> > +            if ( queryRewriteData ) {
> > +                if ( node.type === 'source' ) {
> > +                    var filters = node.filters; // TODO: node.getFilters() when available in camshaft
> > +                    var filters_disabler = _.keys(filters).reduce(
>
> What should be our rule in these cases? always favor ES5 available methods
> over underscore?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/CartoDB/Windshaft-cartodb/pull/449/files/bbb1b4a7b947801709533a8d5871dd9d5db3ffa3#r63696279>
>



-- 
Raul Ochoa
rochoa@cartodb.com

PR #458, opened 2016-05-18
Fix dataview bbox bug when no query rewrite data exists

### LINKED ISSUE #457: Dataviews without query rewrite data and with a bbox fail

jgoizueta:
Fixes #457

### DIFF
--- a/test/acceptance/dataviews/overviews.js
+++ b/test/acceptance/dataviews/overviews.js
@@ -58,6 +58,21 @@ describe('dataviews using tables without overviews', function() {
         });
     });
 
+    it("should admit a bbox", function(done) {
+        var params = {
+            bbox: "-170,-80,170,80"
+        };
+        var testClient = new TestClient(nonOverviewsMapConfig);
+        testClient.getDataview('country_places_count', params, function(err, formula_result) {
+            if (err) {
+                return done(err);
+            }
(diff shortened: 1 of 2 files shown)


jgoizueta:
@rochoa can you CR this?

test/acceptance/dataviews/overviews.js
```diff
@@ -218,16 +250,32 @@ describe('dataviews using tables with overviews', function() {
...
+            }
+            assert.deepEqual(formula_result, {"operation":"sum","result":15,"nulls":0,"type":"formula"});
+
+            testClient.drain(done);
+        });
+    });
+
     describe('filters', function() {
 
         describe('category', function () {
```
rochoa: Why so many changes here?

jgoizueta: The test code was messy (in the layout) and fixed it. I also added test to check dataviews with overviews also don't fail when the bbox parameter is present (this was not failing)

jgoizueta: The one test that detects the problem is in this commit: https://github.com/CartoDB/Windshaft-cartodb/pull/458/commits/5989ab344d450155f71885e8e32d6a6b59b156ac

The rest is just fixing layout/having more coverage for similar cases

Issue #466, opened 2016-05-23
Use Overviews for all Dataviews


jgoizueta:
Extend overviews usage (currently only for formulas and limited to avg, sum, count) to all the dataviews/widgets

rafatower:
can this be closed?

PR #468, opened 2016-05-23
Implement overviews support for all dataview types

jgoizueta:
See #466


jgoizueta:
cc @rochoa can you CR this?

It's been tested in staging, the idea is to further test it in production. Basically overviews are now used for aggregation and histogram dataviews. Cases that cannot be handled well with the current overviews, suc h as minimum/maximum formulas fallback to use the base table.

lib/cartodb/models/dataview/overviews/histogram.js
```diff
@@ -0,0 +1,222 @@
...
+
+var columnTypeQueryTpl = dot.template(
+    'SELECT pg_typeof({{=it.column}})::oid FROM ({{=it.query}}) _cdb_histogram_column_type limit 1'
+);
+var columnCastTpl = dot.template("date_part('epoch', {{=it.column}})");
+
+var BIN_MIN_NUMBER = 6;
+var BIN_MAX_NUMBER = 48;
+
+var basicsQueryTpl = dot.template([
```
rochoa: Can we try to share some of the templates between different dataviews? Some of the query templates are exactly the same and we only change some parts of the CTE.

For instance in this particular case we are just changing a bit the `basics` queries, and the `histogramQueryTpl` query itself. Apart from that everything else is quite similar.

lib/cartodb/models/dataview/overviews/list.js
```diff
@@ -0,0 +1,11 @@
+var BaseOverviewsWidget = require('./base');
```
rochoa: Use `Dataview` for naming.

lib/cartodb/models/dataview/overviews/histogram.js
jgoizueta: Yes, although that might diverge in the feature... I'll give it a thought

lib/cartodb/models/dataview/overviews/list.js
jgoizueta: s/Widget/Dataview

PR #469, opened 2016-05-23
Unify getMapConfig signature in adapters

rochoa:
This de-duplicates steps used in `NamedMapMapConfigProvider` and `MapController`.

Adding new adapters is a matter of following the getMapConfig interface and passing them to the generic adapter constructor, where order matters.

cc @dgaubert please review :)

@jgoizueta this shouldn't not affect you but I guess it's better you are aware of these changes.


lib/cartodb/controllers/map.js
```diff
@@ -134,20 +124,17 @@ MapController.prototype.create = function(req, res, prepareConfigFn) {
     var self = this;
 
     var mapConfig;
-    var analysesResults = [];
+
+    var context = {};
```
rochoa: I don't like this approach too much but I had to use something to share some information between adapters. For instance: overviews adapter consumes some information from analyses adapter.

Probably the best way would be to share through the mapconfig object itself. Any thoughts?

jgoizueta: I also feel uneasy about this, but can't see a solution I like. Either:

* Add it to mapconfig somewhere, but don't think that PG configuration belongs there
* Add a parameter to all step functions to pass along the context from step to step, but this requires either changing functions to which steps are passed as callbacks, or using additional intermediate callbacks... I think this would make the code less clean. And some steps that don't need the context will have to pass it along...
* Current solution, to use a closure variable to share the context. It somehow breaks the encapsulation of each step.

TL;DR; I can't find a better option

dgaubert: After some research, I also didn't find a better approach.

dgaubert:
👍

PR #500, opened 2016-06-13
Adds more error information when either analysis or turbo-carto is not well formed

dgaubert:
cc @rochoa

### DIFF
--- a/lib/cartodb/controllers/base.js
+++ b/lib/cartodb/controllers/base.js
@@ -215,13 +215,9 @@ BaseController.prototype.sendError = function(req, res, err, label) {
     }
 
     var errorResponseBody = {
-        errors: allErrors.map(errorMessage)
+        errors: allErrors.map(errorMessage),
+        errors_with_context: allErrors.map(errorMessageWithContext)
     };
-
-    var errorsWithContext = allErrors.map(errorMessageWithContext).filter(function (err) { return !!err; });
-    if (errorsWithContext.length){
-        errorResponseBody.errors_with_context = errorsWithContext;
-    }
 
     this.send(req, res, errorResponseBody, statusCode);
 };
@@ -240,19 +236,16 @@ function errorMessage(err) {
     var message = (_.isString(err) ? err : err.message) || 'Unknown error';
 
     return stripConnectionInfo(message);
-
 }
 
 function errorMessageWithContext(err) {
-    if (!err.context) {
-        return;
-    }
     // See https://github.com/Vizzuality/Windshaft-cartodb/issues/68
     var message = (_.isString(err) ? err : err.message) || 'Unknown error';
 
     return {
+        type: err.type || 'unknown',
         message: stripConnectionInfo(message),
-        context: err.context
+        context: err.context || 'unknown'
     };
 }
 module.exports.errorMessage = errorMessage;
--- a/lib/cartodb/controllers/base.js
+++ b/lib/cartodb/controllers/base.js
@@ -214,21 +214,46 @@ BaseController.prototype.sendError = function(req, res, err, label) {
         statusCode = 200;
     }
 
-    var errorResponseBody = { errors: allErrors.map(errorMessage) };
+    var errorResponseBody = {
+        errors: allErrors.map(errorMessage)
+    };
+
+    var errorsWithContext = allErrors.map(errorMessageWithContext).filter(function (err) { return !!err; });
+    if (errorsWithContext.length){
+        errorResponseBody.errors_with_context = errorsWithContext;
+    }
 
     this.send(req, res, errorResponseBody, statusCode);
 };
 
-function errorMessage(err) {
-    // See https://github.com/Vizzuality/Windshaft-cartodb/issues/68
-    var message = (_.isString(err) ? err : err.message) || 'Unknown error';
-
+function stripConnectionInfo(message) {
     // Strip connection info, if any
     return message
         // See https://github.com/CartoDB/Windshaft/issues/173
         .replace(/Connection string: '[^']*'\n\s/im, '')
         // See https://travis-ci.org/CartoDB/Windshaft/jobs/20703062#L1644
         .replace(/is the server.*encountered/im, 'encountered');
+}
+
+function errorMessage(err) {
+    // See https://github.com/Vizzuality/Windshaft-cartodb/issues/68
+    var message = (_.isString(err) ? err : err.message) || 'Unknown error';
+
+    return stripConnectionInfo(message);
+
+}
+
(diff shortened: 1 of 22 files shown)


lib/cartodb/controllers/base.js
```diff
@@ -214,22 +214,47 @@ BaseController.prototype.sendError = function(req, res, err, label) {
         statusCode = 200;
     }
 
-    var errorResponseBody = { errors: allErrors.map(errorMessage) };
+    var errorResponseBody = {
+        errors: allErrors.map(errorMessage)
+    };
+
+    var errorsWithContext = allErrors.map(errorMessageWithContext).filter(function (err) { return !!err; });
```
rochoa: I wouldn't filter as you will get different size in both arrays. I would do something like:

```js
{
    errors: allErrors.map(errorMessage),
    context: allErrors.map(errorContext)
}
```

Where:

```js
function errorContext(err) {
    return {
        type: err.type || 'unknown',
        message: errorMessage(err),
        context: err.context // do something with context
    };
}
```

That way what we put in context attribute is a superset of errors. What do you think?

lib/cartodb/models/mapconfig/adapter/turbo-carto-adapter.js
```diff
@@ -86,6 +86,13 @@ TurboCartoAdapter.prototype._parseCartoCss = function (username, layer, callback
         if (err && err.name === 'TurboCartoError') {
             err = new Error('turbo-carto: ' + err.message);
             err.http_status = 400;
+            err.context = {
+                type: 'turbo-carto',
+                layer: {
+                    index: index,
+                    type: layer.type
+                }
```
rochoa: We need to include here info from turbo-carto error itself, that will include the selector/line/column where the problem where found.

rochoa: We need to change turbo-carto to expose the context as an object within the error: https://github.com/CartoDB/turbo-carto/blob/0ad7afc0cc1fd15354e3924659d7a081968c7646/src/helper/turbo-carto-error.js#L3

dgaubert:
it needs that turbo-carto exposes the context as an object

dgaubert:
I think we are ready to merge, don't you?

lib/cartodb/models/mapconfig/adapter/analysis-mapconfig-adapter.js
```diff
@@ -58,13 +58,27 @@ AnalysisMapConfigAdapter.prototype.getMapConfig = function(user, requestMapConfi
 
     requestMapConfig = appendFiltersToNodes(requestMapConfig, dataviewsFiltersBySourceId);
 
-    function createAnalysis(analysisDefinition, done) {
-        self.analysisBackend.create(analysisConfiguration, analysisDefinition, done);
+    function createAnalysis(analysisDefinition, index, done) {
+        self.analysisBackend.create(analysisConfiguration, analysisDefinition, function (err, analysis) {
+            if (err) {
+                err.context = {
+                    type: 'camshaft',
```
rochoa: Not sure about exposing internal naming, wouldn't be better to say `analysis`?

lib/cartodb/models/mapconfig/adapter/turbo-carto-adapter.js
```diff
@@ -84,9 +84,16 @@ TurboCartoAdapter.prototype._parseCartoCss = function (username, layer, callback
     this.turboCartoParser.process(username, layer.options.cartocss, sql, function (err, cartocss) {
         // Only return turbo-carto errors
         if (err && err.name === 'TurboCartoError') {
-            err = new Error('turbo-carto: ' + err.message);
-            err.http_status = 400;
-            return callback(err);
+            var error = new Error('turbo-carto: ' + err.message);
+            error.http_status = 400;
+            error.type = 'turbo-carto';
```
rochoa: Here it's a bit different because this is the external naming as well.

alonsogarciapablo: I'd make this type `layer` (context is actually referring to the layer).

lib/cartodb/models/mapconfig/adapter/turbo-carto-adapter.js
```diff
@@ -84,9 +84,16 @@ TurboCartoAdapter.prototype._parseCartoCss = function (username, layer, callback
...
+            error.context = err.context;
+            error.context.layer = {
+                index: index,
```
alonsogarciapablo: Could we return the `id` of the layer here? I'd love to avoid having to work with layer indexes in CartoDB.js (time has shown that it leads to a lot of bugs). cc: @rochoa @dgaubert

dgaubert: I'm a bit lost here, AFAIK there is no `id` at layer level. I didn't find anything related to an `id` of the layer. Could you provide an example of that?

lib/cartodb/models/mapconfig/adapter/turbo-carto-adapter.js
dgaubert: I'm ok with that, where would it be the right place to locate the `turbo-carto` `context`? for instance:
```
error: {
  type: 'layer',
  layer: {
    index: index,
    context: {
      selector: '...',
      source: '...'
    }
  }
}
```

lib/cartodb/models/mapconfig/adapter/turbo-carto-adapter.js
alonsogarciapablo: I remember @rochoa telling me the same thing in the past. There're probably no ids in Windshaft API and perhaps that made sense in the past, but perhaps now that layers are not the only "resource" that the the API manages, and given that "resources" are related... perhaps it makes sense to start working with ids? (dataviews have ids, analyses have ids...). Thoughts?

dgaubert: I like your idea (not sure about the implications, backwards compatibility and so on), but I don't know if we have time to take care about that.

rochoa: Layers have now an id: https://github.com/CartoDB/Windshaft/blob/e363f7062acd6cff4b7b20c00fa2948e73b4a93c/lib/windshaft/models/mapconfig.js#L105

The id is now returned within the layer metadata: https://github.com/CartoDB/Windshaft/blob/e363f7062acd6cff4b7b20c00fa2948e73b4a93c/lib/windshaft/metadata/layer-metadata.js#L41.

HTH

PR #519, opened 2016-06-21
Now errors with context have the same schema.

dgaubert:
please guys, could you CR this?

cc @rochoa & @alonsogarciapablo

### DIFF
--- a/lib/cartodb/models/mapconfig/adapter/analysis-mapconfig-adapter.js
+++ b/lib/cartodb/models/mapconfig/adapter/analysis-mapconfig-adapter.js
@@ -61,15 +61,15 @@ AnalysisMapConfigAdapter.prototype.getMapConfig = function(user, requestMapConfi
     function createAnalysis(analysisDefinition, index, done) {
         self.analysisBackend.create(analysisConfiguration, analysisDefinition, function (err, analysis) {
             if (err) {
-                err.context = {
-                    type: 'analysis',
-                    analysis: {
-                        index: index,
-                        id: analysisDefinition.id,
-                        type: analysisDefinition.type
-                    }
+                var error = new Error(err.message);
+                error.type = 'analysis';
+                error.context = {};
+                error.context.layer = {
+                    index: index,
+                    id: analysisDefinition.id,
+                    type: analysisDefinition.type
                 };
-                return done(err);
+                return done(error);
             }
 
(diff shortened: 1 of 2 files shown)


alonsogarciapablo:
👍

lib/cartodb/models/mapconfig/adapter/analysis-mapconfig-adapter.js
```diff
@@ -61,15 +61,15 @@ AnalysisMapConfigAdapter.prototype.getMapConfig = function(user, requestMapConfi
...
-                    type: 'analysis',
-                    analysis: {
-                        index: index,
-                        id: analysisDefinition.id,
-                        type: analysisDefinition.type
-                    }
+                var error = new Error(err.message);
+                error.type = 'analysis';
+                error.context = {};
+                error.context.layer = {
```
alonsogarciapablo: @dgaubert why `error.context.layer` and not just `error.context`?

alonsogarciapablo: It's kind of weird cause the `id` is actually the `id` of the analysis, but it's inside of the layer property...

lib/cartodb/models/mapconfig/adapter/analysis-mapconfig-adapter.js
```diff
@@ -61,15 +61,15 @@ AnalysisMapConfigAdapter.prototype.getMapConfig = function(user, requestMapConfi
...
+                    index: index,
+                    id: analysisDefinition.id,
```
rochoa: Be careful, this is the `id` of the analysis node, not the layer id.

alonsogarciapablo: I know... that's why I don't understand why it's inside of `context.layer`...

lib/cartodb/models/mapconfig/adapter/analysis-mapconfig-adapter.js
dgaubert: Right, it does not make sense. I think is better if you provide an example of what do you need for both `analysis` and `layer` (turbo-carto) errors. Because, at the end ,you know better what's needed for this issue.

PR #525, opened 2016-06-26
Fix error context, replaced `turbo-carto` error type by `layer` type

dgaubert:
errors with context, analysis type:
```js
{ 
  errors_with_context: [{
    type: 'analysis',
    message: 'Missing required param "radius"',
    analysis: { id: 'HEAD', type: 'buffer' } 
  }]
}
```
layer type:
```js
{
  errors_with_context: [{
    type: 'layer',
    message: 'turbo-carto: Unable to process "marker-fill". Reason: Unable to compute ramp. Reason: column "wadus_column" does not exist. Context: {"selector":"#populated_places_simple_reduced","source":{"start":{"line":10,"column":3},"end":{"line":10,"column":56}}}',
    http_status: 400,
    layer: {
      index: 0,
      type: 'mapnik',
      context: {
        selector: '#populated_places_simple_reduced',
        source: {
          start: {
            line: 10,
            column: 3
          },
          end: {
            line: 10,
            column: 56
          }
        }
      }
    }
  }]
}
```

please @alonsogarciapablo & @rochoa CR.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,7 @@ Released 2016-mm-dd
 Enhancements:
  - Split turbo-carto adapter substitutions tokens query.
  - Now errors with context have the same schema. #519
+ - Responses with error now returns the layer-id to give more info to the user.
 
 
 ## 2.50.0
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,7 +7,7 @@ Released 2016-mm-dd
 Enhancements:
  - Split turbo-carto adapter substitutions tokens query.
  - Now errors with context have the same schema. #519
- - Responses with error now returns the layer-id to give more info to the user.
+ - Responses with error now return the layer-id to give more info to the user.
 
 
 ## 2.50.0
--- a/lib/cartodb/controllers/base.js
+++ b/lib/cartodb/controllers/base.js
@@ -242,11 +242,19 @@ function errorMessageWithContext(err) {
     // See https://github.com/Vizzuality/Windshaft-cartodb/issues/68
     var message = (_.isString(err) ? err : err.message) || 'Unknown error';
 
-    return {
+    var error = {
         type: err.type || 'unknown',
         message: stripConnectionInfo(message),
-        context: err.context || 'unknown'
     };
+
+    for (var prop in err) {
+        // type & message are properties from Error's prototype and will be skipped
+        if (err.hasOwnProperty(prop)) {
+            error[prop] = err[prop];
+        }
+    }
+
+    return error;
 }
 module.exports.errorMessage = errorMessage;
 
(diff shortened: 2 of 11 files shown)


alonsogarciapablo:
👍 I don't think I'll use the `http_status: 400` and `.layer.context` for now though. I'd merge this so that I can implement the CartoDB.js part, and the we can iterate.

dgaubert:
I kept `http_status` because of backwards compatibility and `.layer.context` could be useful if we want to highlight the error in cartocss editor.

rochoa:
I don't like this one. How do I know it's a turbo-carto/cartocss error?

dgaubert:
Did you mean to use a `type: "analysis"` instead of `turbo-carto`? To me is indifferent, I'll use what you decide. What are your thoughts @alonsogarciapablo?

alonsogarciapablo:
> I don't like this one. How do I know it's a turbo-carto/cartocss error?

This is probably when error codes can come in handy. What do you think?

alonsogarciapablo:
I talked with @rochoa offline about this.

```
{ 
  errors_with_context: [{
    type: 'analysis',
    message: 'Missing required param "radius"',
    analysis: { id: 'HEAD', type: 'buffer' } 
  }]
}
````

☝️ is perfect for now.

```
{
  errors_with_context: [{
    type: 'layer',
    message: 'turbo-carto: Unable...',
    layer: {
      index: 0,
      type: 'mapnik',
      context: {
        selector: '#populated_places_simple_reduced',
        source: {
          start: {
            line: 10,
            column: 3
          },
          end: {
            line: 10,
            column: 56
          }
        }
      }
    }
  }]
}
```

About :point_up: we both agree that `type: 'layer'` is a good idea, but the editor will need to "categorise" these errors a bit more so we might need to add a sub-type somewhere (eg: `turbo-carto`, `sql`, you name it!). I'll think about this a bit more. @dgaubert  suggestions are welcome!

Also, we agreed that it'd be great to replace the `index` by the `id` of the layer, as suggested in a previous comment. We can start changing this!

**Comment by dgaubert** [comment]
that's how it looks right now:

```js
{
  errors_with_context: [{
    type: 'layer',
    subtype: 'turbo-carto',
    message: 'turbo-carto: Unable...',
    layer: {
      id: 'layer0',
      index: 0,
      type: 'mapnik',
      context: {
        selector: '#populated_places_simple_reduced',
        source: {
          start: {
            line: 10,
            column: 3
          },
          end: {
            line: 10,
            column: 56
          }
        }
      }
    }
  }]
}
```
@alonsogarciapablo & @rochoa, that's fine for now, isn't it?

alonsogarciapablo:
Looks good to me! I'd remove `layer.index` since we don't actually need it right now.

rochoa:
I would keep layer index so it covers when you don't send an id for the layer.

alonsogarciapablo:
Ok, but just to be clear, for our current use case (editor + cartodb.js) I'm planning to work with the ids.

PR #543, opened 2016-07-05
Return multiple turbo-carto errors

### LINKED ISSUE #541: Return more multiple errors when map instantiation fails
Right now, when we try to instantiate a map that has multiple errors (eg: two layers with turbo-carto errors and one analysis which is missing a required param), Windshaft-cartodb is only returning the first error.

The following payload, for example, has one turbo-carto error (`wadus` column does not exist) and the analysis is missing the `radius` param):

```
{
  "layers": [
    {
      "id": "ef0fe0d2-e7d2-432b-a603-82a33fe72be5",
      "type": "cartodb",
      "options": {
        "cartocss": "#layer {\n  polygon-fill: ramp([wadus], (#cafdd8, #67cda3, #007553), jenks);\n  line-width: 2;\n  line-color: #FFFFFF;\n  line-opacity: 1;\n}",
        "cartocss_version": "2.1.1",
        "interactivity": [],
        "source": {
          "id": "a1"
        }
      }
    }
  ],
  "dataviews": {},
  "analyses": [
    {
      "id": "a1",
      "type": "buffer",
      "params": {
        "source": {
          "id": "a0",
          "type": "source",
          "params": {
            "query": "SELECT * FROM paradas_metro_madrid_1"
          }
        }
      }
    }
  ]
}
```
Windshaft-CartoDB is returning:

```
{
  "errors": [
    "Missing required param \"radius\""
  ],
  "errors_with_context": [
    {
      "type": "analysis",
      "message": "Missing required param \"radius\"",
      "analysis": {
        "id": "a1",
        "type": "buffer"
      }
    }
  ]
}
```

cc: @dgaubert @rochoa

rochoa:
Closes #541

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,9 @@
 
 Released 2016-mm-dd
 
+Enhancements:
+ - Return multiple turbo-carto errors #541.
+
 Announcements:
  - Upgrades turbo-carto to [0.13.0](https://github.com/CartoDB/turbo-carto/releases/tag/0.13.0).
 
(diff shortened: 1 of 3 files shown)


alonsogarciapablo:
👏

PR #568, opened 2016-08-29
Show & hide support in named maps

### LINKED ISSUE #567: Show/hide support in named maps
Although after instantiating a map it is possible to select the layers you want to render, as per #550, we still miss the option to hide show layers by default at named map level.

The main use case is to have layers hidden on previews.

We need to:
 - [ ] Provide a mechanism at named map level to be able to set the default visibility for different layers.
 - [ ] Decide if that affects layer filters like `all` or `mapnik`, or we want to keep this new feature just for previews.

cc @xavijam

dgaubert:
Fixes #567 

> Provide a mechanism at named map level to be able to set the default visibility for different layers.

User will be able to define in templates which layers are visible. In layer template, for instance:
```
layer_visibility: {
    // "0" = true by default
    "1": false, // layer[index]
    "2": true,
    "layer3": false // using layer.id
}
```
Instances of this template will have layers 0 and 2.

> Decide if that affects layer filters like all or mapnik, or we want to keep this new feature just for previews.

Since instances will have only visible layer, yes. This affects layer filters, `all` and `mapnik` only filter to visible layers.

cc @rochoa


rochoa:
So ... if I set layer_visibility I have no way to instantiate the map with other layer visibility than the default one?

In your example I won't be able to make layer `"layer3"` visible, will I?

dgaubert:
You can always update the named map, can't you?.

BTW my first approach were sending the `layer_visibility` when creates the instance.

I'm not sure what's the best approach.

rochoa:
But if I have to update the named map why I would send all the layers in the first place, in that case, I would send just the layers I want or the ones I'm specifying in the layers_visibility.

The other approach, by itself, is not valid either, because that means that you need to instantiate the map. #567's *main use case is to have layers hidden on previews*.

So either we have something ad-hoc for previews, or we combine this solution with your other approach.

One more thing: does it makes sense to "hide" some layers when I explicitly ask for `all` layers?

dgaubert:
Ooh!

I didn't get it at all. 😓. I'll try to adapt it regarding previews.

docs/named_maps.md
```diff
@@ -73,6 +73,10 @@ The `name` argument defines how to name this "template_name".json. Note that the
       }
     ]
   },
+  "preview_layers": {
```
rochoa: Why don't we make this part of `"view"` and clarify that that property is for previews/static maps?

lib/cartodb/controllers/named_maps.js
```diff
@@ -196,6 +200,46 @@ function numMapper(n) {
     return +n;
 }
 
+function decorateNamedMapProviderWithLayerVisibility (mamedMapProvider) {
```
rochoa: I don't understand this complexity. If this is only for previews: why don't you do the same as we do with other properties in `"view"`?

docs/named_maps.md
dgaubert: :+1:

BTW I'm changing the implementation (again). I think I found the best approach to achieve that.

lib/cartodb/controllers/named_maps.js
dgaubert: Wait, please.

I've just implement another approach. Taking care of final details right now.

rochoa: I'm talking about using a similar approach as in https://github.com/CartoDB/Windshaft-cartodb/blob/377f3d4aff19c7611065cd7fff701a710ff6f5ba/lib/cartodb/controllers/named_maps.js#L143.

dgaubert: Yes, pretty similar.

lib/cartodb/controllers/named_maps.js
```diff
@@ -126,6 +127,23 @@ NamedMapsController.prototype.staticMap = function(req, res) {
         function reqParams() {
             self.req2params(req, this);
         },
+        function getLayerFilterFromPreviewLayers (err) {
```
dgaubert: It's mandatory to do it before getting the provider from cache, thus we can take advantage of cache and created `providers`, `adapters`, `map-config` and so on. The counterpart is we need `template-maps` as dependency in the controller.

rochoa: But the named map provider gives you access to the template, see: https://github.com/CartoDB/Windshaft-cartodb/blob/377f3d4aff19c7611065cd7fff701a710ff6f5ba/lib/cartodb/controllers/named_maps.js#L227-L249.

dgaubert: Yep, but at this point named map provided is already created with req.param.layer equal to 'all' and cache uses that.

https://github.com/CartoDB/Windshaft-cartodb/blob/377f3d4aff19c7611065cd7fff701a710ff6f5ba/lib/cartodb/controllers/named_maps.js#L122

dgaubert: Furthermore `params` is used to create the providerKey for the cache:

https://github.com/CartoDB/Windshaft-cartodb/blob/377f3d4aff19c7611065cd7fff701a710ff6f5ba/lib/cartodb/cache/named_map_provider_cache.js#L25

rochoa: Yes, but params will change, and since it's used as reference the next time you call the cache for the template/provider it will have other value for `params.layer` and will give you a different object.

If you want to make it clearer, you can always retrieve the template, set the `params.layer`, and then recreate the provider by calling self.namedMapProviderCache.get with the modified params object.
IMO this is better because you can understand what is going on, with the other approach you rely on the object being a reference, which will blow our minds in the future if we have a bug here.

I'm pretty sure we don't need to include the template-maps dependency, and we can use the provider for this.

lib/cartodb/controllers/named_maps.js
```diff
@@ -184,6 +202,53 @@ NamedMapsController.prototype.staticMap = function(req, res) {
...
+NamedMapsController.prototype.getLayerFilterFromPreviewLayers = function (cdbUser, template_id, callback) {
+    var self = this;
+
+    step(
+        function getTemplate () {
+            var next = this;
+
+            self.templateMaps.getTemplate(cdbUser, template_id, next);
+        },
+        function getLayerVisivilityFilter (err, template) {
```
rochoa: Everything from here can be done with async functions.

lib/cartodb/controllers/named_maps.js
dgaubert: > If you want to make it clearer, you can always retrieve the template, set the params.layer, and then recreate the provider by calling self.namedMapProviderCache.get with the modified params object.

haha, I tried to implement something like this, and I wasn't able to do it. I'm going to try your approach. Wish me luck.

> with the other approach you rely on the object being a reference, which will blow our minds in the future if we have a bug here.

My mind blew up several times implementing this. 😅 BTW I always avoided to use params as reference, IMHO passing arguments as reference is the worst feature of Javascript  

Thanks for your help.

lib/cartodb/controllers/named_maps.js
rochoa: WITHOUT.

dgaubert:
Shall we merge this?

PR #576, opened 2016-09-23
Pass logger configuration to analysis backend and create a stream

dgaubert:


rochoa:
Two questions:

1. Does this support absolute paths?
1. How does this handle file rotation?

## Example

I set `/tmp/logfile.log` as the log file, I start the app and after a while I do:

```
mv /tmp/logfile.log /tmp/logfile.log.1
```

Will it keep logging to `/tmp/logfile.log.1` or will it be aware of the file change?

dgaubert:
1.  Yes, it creates a stream from the path provided in environment config. It supports both relative and absolute paths.

2. For some reason, I thought bunyan handle it automatically, but I was wrong, we need to do:

```
var log = bunyan.createLogger(...);
...
process.on('SIGUSR2', function () {
    log.reopenFileStreams();
});
```

We have two options:
 1. Camshaft should expose logger and `Windshaft-cartodb` should listen `SIGHUP` event as above.
 2. Camshaft should handle `SIGHUP` event internally.

There is a third way, configuring bunyan to rotate the file itself, w/o logrotate.

I prefer option 2. What do you think??

dgaubert:
As we pass a stream to Camshaft to be used as file destination we can recreate the stream for `SIGHUP` event.

PR #590, opened 2016-11-10
Analyses limit configuration allows to set other limits than timeout

rochoa:
Configuration is now defined as a dictionary instead of just timeouts per analysis type


jgoizueta:
Tested and working with camshaft 0.48.1 :+1:
Thanks! 😄

PR #595, opened 2016-11-23
Upgrade camshaft version to 0.48.2

jgoizueta:
@rochoa can you take a look? Some packages have ben upgraded in the shrinkwrap 😕


rochoa:
:shipit:

PR #604, opened 2016-12-16
Fixes for dataviews using overviews

jgoizueta:
This fixes https://github.com/CartoDB/cartodb/issues/10783


jgoizueta:
@rochoa, can you review this?
There are two different fixes here (corresponding to the two commits):

The first one corrects a defect of the dataview classes for overviews which had an spurious parameters. I'm not exactly sure in which cases a problem was apparent to the user, but probably it caused incorrec results to filtered widgets applied to tables with overviews.

The second fixes this specific problem: CartoDB/cartodb#10783

/cc @rafatower

PR #618, opened 2017-01-17
removed "default" text from Limits descriptions

csobier:
As per a customer Support inquiry, I removed "default" from the Static Maps Limits section, as it was confusing to the user that a default value could be changed.

@rochoa , okay to merge this minor edit?


rochoa:
This is already in master branch. No need to merge it.

PR #624, opened 2017-02-07
Node v6 shrinkwrap

rochoa:


Makefile
```diff
@@ -7,7 +7,13 @@ all:
 	@$(SHELL) ./scripts/install.sh
 
 clean:
-	rm -rf node_modules/*
+	rm -rf node_modules/
+
+shrinkwrap: clean
+	rm npm-shrinkwrap.json
+	npm install --no-shrinkwrap --production
```
dgaubert: Using --production flag, aren't we missing dev dependencies during testing phase in CI?. If I recall correctly, I tried something similar and `npm test` failed because didn't found `mocha`.

rochoa: That's to generate the npm-shrinkwrap.json file, as a shorthand method to not having to type all those commands when regenerating it.

The `devDependencies` will be installed if you don't use `--production` flag along `npm install`, independently of having a npm-shrinkwrap.json file or not.

PR #626, opened 2017-02-17
Camshaft 0.50.0

ethervoid:
Update to [camshaft 0.50.0](https://github.com/CartoDB/camshaft/releases/tag/0.50.0)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ config.status*
 config/environments/*.js
 .idea
 .vscode
+.nvmrc
 tools/munin/windshaft.conf
 logs/
 pids/
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "windshaft-cartodb",
-  "version": "2.87.6",
+  "version": "2.88.0",
   "dependencies": {
     "body-parser": {
       "version": "1.14.2",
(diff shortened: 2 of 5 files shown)


ethervoid:
Please, @dgaubert or @rochoa could you please take a look at this PR. I've deployed in staging and works without any problem but I want to be sure everything is ok.

dgaubert:
About promise issue in `turbo-carto` adapter,  I'm not sure why we weren't able to detect it before. Promises in `turbo-carto` are there since the beginning of `turbo-carto`. @rochoa, do you know why it throws error now?

rochoa:
It is explained in the cherry-picked commit.

package.json
```diff
@@ -20,7 +20,7 @@
   ],
   "dependencies": {
     "body-parser": "~1.14.0",
-    "camshaft": "0.48.5",
+    "camshaft": "0.50.0",
     "cartodb-psql": "~0.6.1",
```
rochoa: I would upgrade cartodb-psql to 0.7.1 here or I would keep this version in the shrinkwrap.

npm-shrinkwrap.json
```diff
@@ -238,27 +238,136 @@
               "resolved": "https://registry.npmjs.org/moment/-/moment-2.17.1.tgz"
             }
           }
+        },
+        "cartodb-psql": {
```
rochoa: As I already pointed, I would upgrade cartodb-psql or remove this. That way we only rely on one version of cartodb-psql.

package.json
ethervoid: Ouch! on it!

dgaubert:
Oh my.. !! I read that comment like three times but I wasn't able to understand it (Friday afternoon effect). Thanks @rochoa and sorry for the noise.

package.json
ethervoid: One question, that means to upgrade [windshaft](https://github.com/CartoDB/Windshaft/blob/master/package.json#L26) dependency too?

.gitignore
```diff
@@ -3,6 +3,7 @@ config.status*
 config/environments/*.js
 .idea
 .vscode
+.nvmrc
```
ethervoid: I've added this to `.gitignore` to avoid upload it by mistake

ethervoid:
Ok, ready to deploy :)

dgaubert:
🇨🇭

Issue #628, opened 2017-02-22
Category "OTHER" is not calculated using the selected agg function


rochoa:
When a category widget displays data aggregated by any functio, the OTHER category still sums the values of the not displayed categories. 

Graphically (aggregating by **AVG(my_column)**):

![image](https://cloud.githubusercontent.com/assets/9017165/23159659/ed5cb092-f823-11e6-9722-dba344c59eee.png)

The expected result should be that the OTHER value should be calculated using the very same aggregating function the widget is using.

In the example above, OTHER should display the average value of the categories not shown in the TOP5, instead of summing? counting? the leftovers

---

Server response:

![image](https://cloud.githubusercontent.com/assets/9017165/23172899/b0eabe90-f857-11e6-90e1-8a47111a2ac4.png)

This supersedes https://github.com/CartoDB/cartodb/issues/11590.

ethervoid:
## Acceptance

1. Create an empty dataset
2. Add this query to make the base of data for the test:
    ```
   WITH data as ( 
        select generate_series(1,3) as val, 'other_a' as cat, ST_SetSRID(ST_MakePoint(0,0),4326) as the_geom, ST_Transform(ST_SetSRID(ST_MakePoint(0,0),4326),3857) as the_geom_webmercator
        UNION ALL
        select generate_series(4,6) as val, 'other_b' as cat, ST_SetSRID(ST_MakePoint(0,0),4326) as the_geom, ST_Transform(ST_SetSRID(ST_MakePoint(0,0),4326),3857) as the_geom_webmercator
        UNION ALL
        select generate_series(7,9) as val, 'other_c' as cat, ST_SetSRID(ST_MakePoint(0,0),4326) as the_geom, ST_Transform(ST_SetSRID(ST_MakePoint(0,0),4326),3857) as the_geom_webmercator
        UNION ALL
        select generate_series(10,12) as val, md5(generate_series(0,4)::text) as cat, ST_SetSRID(ST_MakePoint(0,0),4326) as the_geom, ST_Transform(ST_SetSRID(ST_MakePoint(0,0),4326),3857) as the_geom_webmercator
    )
    SELECT row_number() OVER (ORDER BY val asc) as cartodb_id, val, cat, the_geom, the_geom_webmercator FROM data
    ```
3. Create a map using the `Create Map` button in the table view
3. Create a new category widget using the cat column and:
    - [ ] Check COUNT operation returns `9`
[...]

PR #633, opened 2017-02-27
Other category now uses the selected aggregated function

ethervoid:
Related #628

### DIFF
--- a/test/acceptance/dataviews/aggregation.js
+++ b/test/acceptance/dataviews/aggregation.js
@@ -1,5 +1,4 @@
 require('../../support/test_helper');
-var _ = require('underscore');
 var assert = require('../../support/assert');
 var TestClient = require('../../support/test-client');
 
@@ -117,7 +116,7 @@ describe('aggregations happy cases', function() {
         'select generate_series(10,12) as val, md5(generate_series(0,4)::text) as cat, NULL as the_geom_webmercator'
     ].join(' UNION ALL ');
 
-    _.keys(operations_and_values).forEach(function (operation) {
+    Object.keys(operations_and_values).forEach(function (operation) {
         var description = 'should aggregate OTHER category using "' + operation + '"';
 
         it(description, function (done) {
@@ -132,9 +131,11 @@ describe('aggregations happy cases', function() {
                 assert.equal(aggregation.count, 24);
                 assert.equal(aggregation.nulls, 0);
 
-                var other_category = _.find(aggregation.categories,
-                                            function(category){ return category.category === 'Other'; });
-                assert.equal(other_category.value, operations_and_values[operation]);
+                var aggregated_categories = aggregation.categories.filter( function(category) {
+                    return category.agg === true;
+                });
+                assert.equal(aggregated_categories.length, 1);
+                assert.equal(aggregated_categories.shift().value, operations_and_values[operation]);
 
                 done();
             });
--- a/test/acceptance/dataviews/aggregation.js
+++ b/test/acceptance/dataviews/aggregation.js
@@ -1,5 +1,5 @@
 require('../../support/test_helper');
-
+var _ = require('underscore');
 var assert = require('../../support/assert');
 var TestClient = require('../../support/test-client');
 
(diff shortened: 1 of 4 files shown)


ethervoid:
Please @rochoa could you CR this whenever is possible? Thank you!

test/acceptance/dataviews/aggregation.js
```diff
@@ -1,5 +1,5 @@
 require('../../support/test_helper');
-
+var _ = require('underscore');
```
rochoa: I would avoid underscore as much as possible.

test/acceptance/dataviews/aggregation.js
```diff
@@ -107,4 +107,37 @@ describe('aggregations happy cases', function() {
...
+    var operations_and_values = {'count': 9, 'sum': 45, 'avg': 5, 'max': 9, 'min': 1};
+
+    var query_other = [
+        'select generate_series(1,3) as val, \'other_a\' as cat, NULL as the_geom_webmercator',
+        'select generate_series(4,6) as val, \'other_b\' as cat, NULL as the_geom_webmercator',
+        'select generate_series(7,9) as val, \'other_c\' as cat, NULL as the_geom_webmercator',
+        'select generate_series(10,12) as val, md5(generate_series(0,4)::text) as cat, NULL as the_geom_webmercator'
+    ].join(' UNION ALL ');
+
+    _.keys(operations_and_values).forEach(function (operation) {
```
rochoa: You can use `Object.keys(operations_and_values).forEach` as an alternative.

test/acceptance/dataviews/aggregation.js
```diff
@@ -107,4 +107,37 @@ describe('aggregations happy cases', function() {
...
+                assert.ifError(err);
+
+                assert.ok(aggregation);
+                assert.equal(aggregation.type, 'aggregation');
+                assert.ok(aggregation.categories);
+                assert.equal(aggregation.categoriesCount, 8);
+                assert.equal(aggregation.count, 24);
+                assert.equal(aggregation.nulls, 0);
+
+                var other_category = _.find(aggregation.categories,
```
rochoa: What about looking for the `agg: true` entry? Which ensures is the aggregated category. That way you can also ensure in the test that there is just one aggregated category.

test/acceptance/dataviews/aggregation.js
```diff
@@ -107,4 +107,37 @@ describe('aggregations happy cases', function() {
...
         });
     });
+
+    var operations_and_values = {'count': 9, 'sum': 45, 'avg': 5, 'max': 9, 'min': 1};
+
+    var query_other = [
+        'select generate_series(1,3) as val, \'other_a\' as cat, NULL as the_geom_webmercator',
+        'select generate_series(4,6) as val, \'other_b\' as cat, NULL as the_geom_webmercator',
+        'select generate_series(7,9) as val, \'other_c\' as cat, NULL as the_geom_webmercator',
+        'select generate_series(10,12) as val, md5(generate_series(0,4)::text) as cat, NULL as the_geom_webmercator'
```
rochoa: What about using a different query here, it's a bit hard to reason about the double generate_series, where the first one loops over 10,11,12 vals.

test/acceptance/dataviews/aggregation.js
ethervoid: Oh, ok. This is to avoid having dependency of external libraries for the test part?

rochoa: Not that much about testing, but about relying more on language semantics, in future versions we will be able to use Object.values and Object.entries, which will make it more clear.

test/acceptance/dataviews/aggregation.js
ethervoid: That was my first code but with underscore looks more readable. Now I'll change to it without underscore

test/acceptance/dataviews/aggregation.js
ethervoid: Good point

test/acceptance/dataviews/aggregation.js
ethervoid: I was looking for a query with the same data, therefore I'm using generate_series. But the order by value-rank hits me when I was trying to make a more readable query.

I wasn't sure if it was going to be hard to read, I thought it was not because I put the objective categories as `other_*` but it seems to be hard to understand. I have to find a better approach.

rochoa: For me, the confusing part is just the last one, especially the md5 with the generate_series. Once you look at what it returns you can pretty much understand it, however, I think that will be difficult to understand for the future ourselves.

What about something like:

```sql
postgres=# select unnest(ARRAY[10,20,30,40,50]) as val, unnest(ARRAY['a','b','c','d','e']) as cat, NULL as the_geom_webmercator from generate_series(1,3);
 val | cat | the_geom_webmercator
-----+-----+----------------------
  10 | a   |
  20 | b   |
  30 | c   |
  40 | d   |
  50 | e   |
  10 | a   |
  20 | b   |
  30 | c   |
  40 | d   |
  50 | e   |
  10 | a   |
  20 | b   |
  30 | c   |
  40 | d   |
  50 | e   |
(15 rows)
```

You still have 5 categories, with enough hard-coded values for each of them in order to keep the `other_*` as Other (agg:true).

rochoa: BTW, that's similar to:

```sql
select generate_series(10,50,10) as val, unnest(ARRAY['a','b','c','d','e']) as cat, NULL as the_geom_webmercator from generate_series(1,3);
```

rochoa: Oh, I see you already approached the situation with another solution 👍 .

ethervoid: Yes, I thought about add `unnest` and so on but for 5 categories I think is best to have it easy to understand. In case we need more categories, we could change it easily.

Thank you very much for your solutions!

test/acceptance/dataviews/aggregation.js
```diff
@@ -107,4 +106,43 @@ describe('aggregations happy cases', function() {
...
+                assert.ok(aggregation.categories);
+                assert.equal(aggregation.categoriesCount, 8);
+                assert.equal(aggregation.count, 24);
+                assert.equal(aggregation.nulls, 0);
+
+                var aggregated_categories = aggregation.categories.filter( function(category) {
+                    return category.agg === true;
+                });
+                assert.equal(aggregated_categories.length, 1);
+                assert.equal(aggregated_categories.shift().value, operations_and_values[operation]);
```
rochoa: This is a super small detail. But I usually avoid modifying arrays/objects when validating/asserting. And as you have already guaranteed the array has its size equals to 1, I would assert against `aggregated_categories[0].value`.

I know it's not important in this case, but I always try to avoid modifying arrays/objects when I'm not the `owner`. yes, you could argue that in this case, you are the owner.

rochoa:
👍  All my comments were nice things to have/improve. So thank you for your patience.

We are good to :shipit:.

test/acceptance/dataviews/aggregation.js
ethervoid: Makes sense, changing

Issue #634, opened 2017-03-02
Category "OTHER" is not calculated using the selected agg function in OVERVIEWS


ethervoid:
Related #628 #625 

This was fixed for the non-overviews dataviews but is not working as intended for overviews dataviews

jgoizueta:
As @rochoa has commented we may have to reconsider how this is implemented, but for now I think we could review the state of possible dataviews/overviews inconstancies and fix the problems in  #634 and #625 together with any other we detect in the review.

Maybe we should rename this ticket to take care of it all here, what do you think?

PR #637, opened 2017-03-13
Histogram column type discovery query uses non-filtered query

### LINKED ISSUE #606: Histogram column type discovery should use original query
Histograms try to discover the column type to allow transparent numeric and date columns.

The query to discover the column type uses the filtered query (using the bounding box filter, for instance), in case no data is returned, the column is defaulted to the ::numeric type, and it fails when the column is a date.

rochoa:
Fixes #606.

### DIFF
--- a/lib/cartodb/models/dataview/factory.js
+++ b/lib/cartodb/models/dataview/factory.js
@@ -11,7 +11,7 @@ var DataviewFactory = {
         if (!this.dataviews[type]) {
             throw new Error('Invalid dataview type: "' + type + '"');
         }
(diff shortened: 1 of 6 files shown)


dgaubert:
Do we have to apply the same to `overviews`???

rochoa:
Short answer: yes.

There is an issue for everything "overviews", see https://github.com/CartoDB/Windshaft-cartodb/issues/634.

I guess this case is easier to solve and we could apply the same approach to overviews. Let me try.

dgaubert:
CR: 🇪🇸

Issue #629, opened 2017-02-22
Problems with analyses cache headers and invalidations


rochoa:
## Prerequisite

1. Use Builder to simplify the STR.
1. Have a points dataset `s1`.

## STR

1. Create a Map with our `s1` points dataset.
1. Add a new analysis: "Create areas of influence", use default params.
   - This creates a temporary table for the analysis, let's say `b1`.
1. Go to the embed and check the headers for the Maps API Named Map instantiation:
   - X-Cache-Channel: will be `b1`, from the analysis table.
   - Surrogate-Key: will contain the named map key, `n:{hash(name)}`; but also the table key: `t:{hash(b1)}`.
1. Run, to simulate the dataset changed, the query:
```sql
select cdb_tablemetadatatouch('s1');
```
5. Reload the named map instantiation and check the headers again:
   - X-Cache-Channel: will be `b2`, as the source table changed, the analysis table is new and has a different name.
   - Surrogate-Key: the same happens with the table key: `t:{hash(b2)}`.

## Problems with this approach

This, per se, it's not a problem because the headers have the correct headers.

However, this becomes a problem when there is a cache layer relying on those headers for its invalidation mechanism. Actually, there are several problems:

1. The named map response never contains any header associated with the original `s1` table. Once that table changes (in our scenario due to `cdb_tablemetadatatouch`), we don't invalidate the response because it was never tag/annotated with any information about the original `s1` table.
1. As analyses tables are not cartodbifyed, they won't trigger any invalidation. In our case that means that once `s1` is modified and the analysis runs again, we won't invalidate any object tagged with `b1` automatically. As an alternative, we could manually trigger an invalidation for `b1` once the analysis is computed again, but at that point, the analysis is totally different because `s1` already changed, and we will only know about `b2` and nothing about `b1` anymore.
1. Another alternative is to tag the named map with the original `s1` table name. That way, we will invalidate the named map when the data at `s1` changes. However, that introduces a race condition, because if the named map is instantiated again before the analysis for `b2` finishes, we will cache it again before the analysis finished.

## Potential solution

1. Tag the named map response with the original `s1` table name.
1. Each analysis invalidates all objects associated to the source tables they rely on after the analysis is finished, avoiding the race condition.

The problem with this approach is we will invalidate other objects associated to the original `s1` table, and not only the named maps associated to analyses relying on it.

rochoa:
Related issue: https://github.com/CartoDB/cartodb/issues/11157.

[...]

PR #638, opened 2017-03-13
Fix cache invalidation problem with affected tables

ethervoid:
Related #629

### DIFF
--- a/lib/cartodb/controllers/layergroup.js
+++ b/lib/cartodb/controllers/layergroup.js
@@ -391,13 +391,15 @@ LayergroupController.prototype.getAffectedTables = function(user, dbName, layerg
                 function getSQL(err, mapConfig) {
                     assert.ifError(err);
 
-                    var queries = mapConfig.getLayers()
-                        .map(function(lyr) {
-                            return lyr.options.sql;
-                        })
-                        .filter(function(sql) {
-                            return !!sql;
-                        });
+                    var queries = [];
+                    mapConfig.getLayers().map(function(layer) {
+                        queries.push(layer.options.sql);
+                        if (layer.options.affected_tables) {
+                            layer.options.affected_tables.map(function(table) {
+                                queries.push('SELECT * FROM ' + table + ' LIMIT 0');
+                            });
+                        }
+                    });
 
                     return queries.length ? queries.join(';') : null;
                 },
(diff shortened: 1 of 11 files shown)


node-windshaft.log
```diff
@@ -0,0 +1,145 @@
+{"name":"camshaft","hostname":"cdb-dev","pid":21600,"level":30,"analysis":"9de00c87a82f0929cfc345d2d9f1d01e05811bc4","node":"9de00c87a82f0929cfc345d2d9f1d01e05811bc4","type":"buffer","queued":true,"username":"localhost","msg":"analysis:node","time":"2017-03-14T09:14:16.981Z","v":0}
```
ethervoid: what file :trollface:

lib/cartodb/controllers/layergroup.js
```diff
@@ -388,13 +388,15 @@ LayergroupController.prototype.getAffectedTables = function(user, dbName, layerg
...
 
-                    var queries = mapConfig.getLayers()
-                        .map(function(lyr) {
-                            return lyr.options.sql;
-                        })
-                        .filter(function(sql) {
-                            return !!sql;
-                        });
+                    var queries = [];
+                    mapConfig.getLayers().forEach(function(layer) {
```
dgaubert: do you think it is worth trying to create a method in MapConfig that returns all affected tables from every layer?. If so, we can use it in `controllers/map.js` too.

ethervoid: I don't think it worths it because is a very specific use case. I don't know if I'm wrong but the regular case will be to ask for the affected tables for a layer which is not the case.

What do you think?

dgaubert: Fine by me, but I'm having the feel that everything is specific and nothing can be reused. That's another issue about how we are designing the whole system. In any case, it's just my own fear and nothing related to this PR.

Sorry for the noise. 😢

ethervoid: No problem at all, for me it's a totally legit concern and you've by far more knowledge about how this system is made so if you think this change should be made then I should move part of this logic to Mapconfig model to avoid duplication.

One think that makes me suspicious is [this filter](https://github.com/CartoDB/Windshaft-cartodb/pull/638/files/4371360fc2c0cb8e4698790fd671225de3b16b59#diff-e3036db619c9174d944dc5d7e9992c80L396) I've removed. It's possible to have layers without SQL?. If the answer is yes, I should change this method to reflect that

dgaubert: No, no. Let's move it ahead!

> It's possible to have layers without SQL?

Following the [docs](https://github.com/CartoDB/Windshaft/blob/master/doc/MapConfig-1.5.0.md), the answer is: no it isn't.

PR #652, opened 2017-03-30
Update windshaft to 3.1.0

rochoa:


yarn.lock
```diff
@@ -781,13 +781,6 @@ glob@^7.0.5:
     once "^1.3.0"
     path-is-absolute "^1.0.0"
 
-"glob@~ 3.2.1":
```
dgaubert: I added it last time that I ran `yarn upgrade`. I guess, we have to run `yarn prune` or equivalent before updating `yarn.lock`

yarn.lock
```diff
@@ -1231,13 +1224,6 @@ mime@~1.2.11:
   version "1.2.11"
   resolved "https://registry.yarnpkg.com/mime/-/mime-1.2.11.tgz#58203eed86e3a5ef17aed2b7d9ebd47f0a60dd10"
 
-minimatch@0.3:
```
dgaubert: Same here :(

PR #655, opened 2017-03-30
Static maps layers filter

### LINKED ISSUE #653: Static map endpoints should allow to specify the layers to render

rochoa:
This closes #653 .


lib/cartodb/middleware/allow-query-params.js
```diff
@@ -0,0 +1,6 @@
+module.exports = function allowQueryParams(params) {
+    return function allowQueryParamsMiddleware(req, res, next) {
```
dgaubert: Could we ensure here that params is an Array instead of checking it in `req2params` function?

rochoa: If the middleware is not used you still need to check if it's an array or not. However, it would be nice to check it here and throw and error if used incorrectly.

PR #673, opened 2017-04-24
[windshaft realease required] Manage multiple values of buffer-size for different formats

### LINKED ISSUE #665: Manage multiple values of buffer-size for different formats
We're going to need to manage different values of `buffer-size` for different formats so we need to make some changes:

#### Tasks

- [ ] Should be able to receive different configurations for the buffer-size for environment 
        configuration: dictionary with different `buffer-size` per format
- [ ] Create renderer using proper `buffer-size` for the requested format
- [ ] Named maps should have a template for the `buffer-size` property so the user could create multiple instantiations with different `buffer-size`

dgaubert:
Fixes #665


lib/cartodb/backends/template_maps.js
```diff
@@ -474,6 +478,15 @@ TemplateMaps.prototype.instance = function(template, params) {
 
   // NOTE: we're deep-cloning the layergroup here
   var layergroup = JSON.parse(JSON.stringify(template.layergroup));
+
+  if (typeof layergroup.buffersize === 'string') {
```
rochoa: Although this is a nice way to do change the buffer size, it requires changing all existing named maps to have the placeholders.

Furthermore, I see this as a specific use case, not a general purpose one. I mean, we are exposing a way to change the buffer size per format, so I don't think we should give the user the responsibility of creating the template with a placeholder to change later the buffer size. Did I explain myself?

dgaubert: Hey @rochoa, thanks for reviewing this. 

> Although this is a nice way to do change the buffer size, it requires changing all existing named maps to have the placeholders.

I don't see the reason to change all existing named-maps. All of them (with no buffer-size defined) will work with the default value of the renderer through options. So `mvt` will use default value set to `0` and raster will use `64`. Do we forget something?

dgaubert: >  I don't think we should give the user the responsibility of creating the template with a placeholder to change later the buffer size.

it's totally optional, It will work for the 90% of the users (using default values from options) and, for instance, it's a simple way to `mobile` define the buffer-size that they need with named maps.

lib/cartodb/controllers/layergroup.js
```diff
@@ -56,7 +56,7 @@ LayergroupController.prototype.register = function(app) {
         this.tile.bind(this));
 
     app.get(app.base_url_mapconfig +
-        '/:token/:z/:x/:y.:format', cors(), userMiddleware,
+        '/:token/:z/:x/:y.(:format)', cors(), userMiddleware,
```
rochoa: I would like to know why we needed this change.

lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
```diff
@@ -0,0 +1,35 @@
...
+module.exports = MapConfigNamedMapAdapter;
+
+MapConfigNamedMapAdapter.prototype.getMapConfig = function (user, requestMapConfig, params, context, callback) {
+    if (context.templateParams && 
+            context.templateParams.buffersize && 
+            isValidBufferSize(context.templateParams.buffersize)) {
+        requestMapConfig.buffersize = context.templateParams.buffersize;
+    }
+
+    process.nextTick(function () {
```
rochoa: Any particular reason to use nextTick here?

lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
```diff
@@ -0,0 +1,35 @@
...
+
+function MapConfigNamedMapAdapter() {
+}
+
+module.exports = MapConfigNamedMapAdapter;
+
+MapConfigNamedMapAdapter.prototype.getMapConfig = function (user, requestMapConfig, params, context, callback) {
+    if (context.templateParams && 
+            context.templateParams.buffersize && 
+            isValidBufferSize(context.templateParams.buffersize)) {
```
rochoa: This approach is a bit _all or nothing_, if for some reason the user sends `{buffersize: {png: 0, wadus: 32}}`, we will skip the png value. Furthermore, if the user already created a named map with buffersize like: `{mvt: 0}` and later tries to override just the png buffer size with `{buffersize: {png: 32}}` it will destrou the mvt one.

What about something like:

```diff
diff --git a/lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js b/lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
index 76e8f3f5..970e3ee2 100644
--- a/lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
+++ b/lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
@@ -5,31 +5,20 @@ function MapConfigNamedMapAdapter() {

 module.exports = MapConfigNamedMapAdapter;

+var formats = ['png', 'png32', 'mvt', 'grid.json', 'geojson'];
+
 MapConfigNamedMapAdapter.prototype.getMapConfig = function (user, requestMapConfig, params, context, callback) {
-    if (context.templateParams &&
-            context.templateParams.buffersize &&
-            isValidBufferSize(context.templateParams.buffersize)) {
-        requestMapConfig.buffersize = context.templateParams.buffersize;
+    if (!context.templateParams.buffersize) {
+        return callback(null, requestMapConfig);
     }

+    formats.forEach(function(format) {
+        if (Number.isFinite(context.templateParams.buffersize[format])) {
+            requestMapConfig.buffersize[format] = context.templateParams.buffersize[format];
+        }
+    });
+
     process.nextTick(function () {
         callback(null, requestMapConfig);
     });
 };
-
-function isValidBufferSize (bufferSize) {
-    var formats = ['png', 'png32', 'mvt', 'grid.json', 'geojson'];
-
-    if (!_.isObject(bufferSize) || (_.isArray(bufferSize) || _.isFunction(bufferSize))) {
-        return false;
-    }
-
-    for (var index = 0; index < formats.length; index++) {
-        var bufferSizeByFormat = bufferSize[formats[index]];
-        if (bufferSizeByFormat && !Number.isFinite(bufferSizeByFormat)) {
-            return false;
-        }
-    }
-
-    return true;
-}
```

lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
```diff
@@ -0,0 +1,35 @@
+var _ = require('underscore');
+
+function MapConfigNamedMapAdapter() {
+}
+
+module.exports = MapConfigNamedMapAdapter;
+
+MapConfigNamedMapAdapter.prototype.getMapConfig = function (user, requestMapConfig, params, context, callback) {
```
rochoa: Not sure about the naming, what about building something `buffersize` specific.

lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
```diff
@@ -0,0 +1,35 @@
...
+    if (context.templateParams && 
+            context.templateParams.buffersize &&
```
rochoa: Same about this, what about not exposing the whole `templateParams` in the context and just the buffer size?

lib/cartodb/models/mapconfig/provider/named-map-provider.js
```diff
@@ -90,6 +90,7 @@ NamedMapMapConfigProvider.prototype.getMapConfig = function(callback) {
         },
         function instantiateTemplate(err, templateParams) {
             assert.ifError(err);
+            context.templateParams = templateParams;
```
rochoa: I had some concerns about this, but after reviewing the whole situation I see what you did here. It's not easy to access the request `configuration` without parsing again the query and body params, not ideal in the part of the stack where we should work with the buffer size.

So, FAIR ENOUGH.

lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
```diff
@@ -0,0 +1,35 @@
...
+        requestMapConfig.buffersize = context.templateParams.buffersize;
+    }
+
+    process.nextTick(function () {
+        callback(null, requestMapConfig);
+    });
+};
+
+function isValidBufferSize (bufferSize) {
+    var formats = ['png', 'png32', 'mvt', 'grid.json', 'geojson'];
```
rochoa: I don't know if at this point we should have all supported formats centralized at some common place. On the other hand, I would not add `geojson` to this equation. We should drop `geojson` support sooner than later.

lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
dgaubert: Yep, there is a reason. As you already know `getMapConfig` is a function that "implements" an implicit interface that is the same for any map-config adapter. That function receives a callback and other adapters do work asynchronously. In order to keep that expectation and to avoid [releasing Zalgo](http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony) we use this synthetic deferral (nextTick).

lib/cartodb/controllers/layergroup.js
dgaubert: One [test](https://github.com/CartoDB/Windshaft-cartodb/pull/673/files#diff-b352fef23a080c1374a7f2a2e86c185eR128) failed because of this. For this path: `/api/v1/map/a924d7138f2d5d880e70eea5a8de987d:0/7/64/48.grid.json`  we get the following params: 
```
{ 
  z: '7',
  x: '64',
  y: '48.grid',
  format: 'json' 
}
```
returning 400:
```
{
    "errors": [
        "Unsupported format json"
    ],
    "errors_with_context": [
        {
            "type": "unknown",
            "message": "Unsupported format json"
        }
    ]
}
```

I guess it is an issue in express but using parenthesis was fixed somehow.

lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
dgaubert: 👍

lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
dgaubert: Further info: https://www.quora.com/What-is-Zalgo-in-the-context-of-Node-js

lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
dgaubert: Applied, really nice. thanks

lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
dgaubert: We named it a bit generic in order to add some further adaptation. But you are right, at the moment something specific is better. What about `MapConfigBufferSizeAdapter` ?

lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
dgaubert: after this [comment](https://github.com/CartoDB/Windshaft-cartodb/pull/673#discussion_r115536587) is no longer valid, is it?

lib/cartodb/controllers/layergroup.js
rochoa: That's because it is missing the layer. Grid.json only supports `/:layer/:z/:x/:y.grid.json` URL schema.

lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
rochoa: 👍.

lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
rochoa: 👌 😈.

lib/cartodb/controllers/layergroup.js
dgaubert: Oh my God, I've checked that url many times and always saw `:0` as layer index... 😱  

Fixing it...

dgaubert: That was my bad, already done!

Anything else? 

Thanks for reviewing this.

lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
dgaubert: I've tried to move it to a common place but I didn't see a good solution for that. There is just another place where we have to deal with formats in `windshaft-cartodb` and it's just to build the suffix to send metrics and that includes torque formats which we don't take into account for `buffer-size`customization. At the end, I prefer to keep formats as close as possible to `buffer-size-adapter`. What do you think?

On the other hand, I removed `geojson` support for `buffer-size` customization.

lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
rochoa: Really nice piece of information 👍 .

lib/cartodb/models/mapconfig/adapter/mapconfig-named-map-adapter.js
rochoa: Let's keep them here. If we are not sure yet, and as rule of thumb, I prefer the duplication over the wrong abstraction 😈 .

PR #683, opened 2017-05-16
Use windshaft-stats to extract info about layer metadata

### LINKED ISSUE #664: Include stats in the response to map instantiation requests
Required for https://github.com/CartoDB/cartodb/issues/11783.

For now, we just need to know the number of vertexes of all the layers in the "layer group".

cc: @rochoa

dgaubert:
Fixes #664


lib/cartodb/backends/layer-stats/empty-layer-stats.js
```diff
@@ -0,0 +1,16 @@
...
+    this._types = types || {};
+}
+
+EmptyLayerStats.prototype.is = function (type) {
+    return this._types[type] ? this._types[type] : false;
+};
+
+EmptyLayerStats.prototype.getStats =
+function (layer, dbConnection, callback) {
+    process.nextTick(function() {
```
ethervoid: Change this for setImmediate?

lib/cartodb/backends/layer-stats/layer-stats.js
```diff
@@ -0,0 +1,45 @@
+var d3 = require('d3-queue');
```
ethervoid: We should change this for `queue-async` or move all the `queue-async` to `d3-queue` either in `package.json` and all the files we use it.

lib/cartodb/backends/layer-stats/mapnik-layer-stats.js
```diff
@@ -0,0 +1,28 @@
+var queryUtils = require('./query-utils');
```
ethervoid: Maybe we could move this to lib/cartodb/query-utils.js

lib/cartodb/controllers/map.js
```diff
@@ -238,8 +244,9 @@ MapController.prototype.instantiateTemplate = function(req, res, prepareParamsFn
             );
             mapConfigProvider.getMapConfig(this);
         },
-        function createLayergroup(err, mapConfig_, rendererParams) {
+        function createLayergroup(err, mapConfig_, _rendererParams) {
```
ethervoid: _rendererParams is not used anymore for stats

lib/cartodb/backends/stats.js
```diff
@@ -0,0 +1,16 @@
...
+
+module.exports = StatsBackend;
+
+StatsBackend.prototype.getStats = function(mapConfig, dbConnection, callback) {
+    var enabledFeatures = global.environment.enabledFeatures;
+    var layerStatsEnabled = enabledFeatures ? enabledFeatures.layerStats: false;
+    if (layerStatsEnabled) {
+        layerStats().getStats(mapConfig, dbConnection, callback);
+    } else {
+        return callback(null, null);
```
rochoa: Should we return an object instead of null? So we return the same collection type (whether is an object or an array) and we don't need to check for the null response outside.

ethervoid: Fair point! changing

ethervoid: Changed. I keep checking for returned data because I don't want to show the stats property if the FF is disabled.

PR #698, opened 2017-06-02
Support histograms with date/time expressions for aggregations

### LINKED ISSUE #691: [time-series] Support histograms with date/time expressions for aggregations
Since dates are represented internally as numeric values, it is possible to use the current histogram on dates as well, though accuracy will be compromised. The reason for this is in the fact that time based intervals are not fixed (think of leap years and on the number of days in a month). 

So we need special support for histograms with date column types and perform aggregations based on date/time expressions, such as: 
 - `minute `
 - `hour`
 - `day`
 - `week`
 - `month`
 - `quarter`
 - `year`

cc @jorgesancha, @rafatower, @rochoa

dgaubert:
Fixes #691


lib/cartodb/models/dataview/histogram.js
```diff
@@ -105,7 +180,8 @@ var TYPE = 'histogram';
      type: 'histogram',
      options: {
          column: 'name',
-         bins: 10 // OPTIONAL
+         bins: 10, // OPTIONAL
+         aggregation: 'day' // OPTIONAL
```
ethervoid: `timezone` parameter?

lib/cartodb/models/dataview/histogram.js
```diff
@@ -249,7 +389,11 @@ Histogram.prototype.format = function(result, override) {
         width = firstRow.bin_width || width;
         avg = firstRow.avg_val;
         nulls = firstRow.nulls_count;
-        binsStart = override.hasOwnProperty('start') ? getBinStart(override) : firstRow.min;
+        binsStart = override.hasOwnProperty('start') ?
```
ethervoid: This is hard to read/understand. Maybe move it to a method?

dgaubert: 🆗

lib/cartodb/models/dataview/histogram.js
```diff
@@ -97,6 +97,81 @@ var histogramQueryTpl = dot.template([
...
+    '    SELECT',
+    '        bins_array,',
+    '        ARRAY_LENGTH(bins_array, 1) AS bins_number',
+    '    FROM (',
+    '       SELECT',
+    '           ARRAY(',
+    '               {{?it._aggregation==="quarter"}}',
+    '               SELECT GENERATE_SERIES(start_date, end_date, \'3 month\'::interval)',
+    '               {{??}}',
+    '               SELECT GENERATE_SERIES(start_date, end_date, \'1 {{=it._aggregation}}\'::interval)',
```
ethervoid: We've to be aware that a user aggregating, in the worst case, by minute for a 10 year period could lead to 5M rows. Maybe we should add some limits here?

jorgesancha: @saleiva @noguerol this is an important point - there is a technical side to this (what kind of limits we want to impose and what should happen if they are exceeded) and a product side (whether we should even allow certain types of aggregations - minutes, hours - if the data is huge)

jorgesancha: Maybe we could address this ^^^ in a separate ticket

saleiva: Sure, I think we should limit depending on the "total duration" of the data, for sure.

lib/cartodb/models/dataview/histogram.js
```diff
@@ -97,16 +97,102 @@ var histogramQueryTpl = dot.template([
     'ORDER BY bin'
 ].join('\n'));
 
+var dateBasicsQueryTpl = dot.template([
+    'basics AS (',
+    '    SELECT',
+    '        max(date_part(\'epoch\', {{=it._column}})) AS max_val,',
```
ethervoid: This is going to fail. You have to cast it to numeric date_part(....)::numeric if not you're going to get:
```
function avg(timestamp with time zone) does not exist
```

lib/cartodb/models/dataview/histogram.js
```diff
@@ -97,16 +97,102 @@ var histogramQueryTpl = dot.template([
...
+    '        min(date_part(\'epoch\', {{=it._column}})) AS min_val,',
+    '        avg(date_part(\'epoch\', {{=it._column}})) AS avg_val,',
```
ethervoid: This is going to fail in the `avg` which doesn't support date/time. You have to cast it to numeric date_part(....)::numeric if not you're going to get:
```
function avg(timestamp with time zone) does not exist
```

dgaubert: Did you get this error? I've tested this query intensively and I didn't get this error

dgaubert: I meant, `date_part` from `epoch` returns a numeric value: seconds since `epoch`. Which is the same approach that we have so far. See line 11.

ethervoid: Yes, it's working, my bad. 

Because I didn't pass the api_key when I called the dataview url and it tries to get the data from the private table with a `publicuser` which l pass as parameter. So it fails to get the type of column and try to execute the sql for non-date data.

Maybe we could take a looks at this and check the result [here](https://github.com/CartoDB/Windshaft-cartodb/blob/d3bcf6f80d4d1d2802609a55116c796e4478fe9d/lib/cartodb/models/dataview/histogram.js#L231)

lib/cartodb/models/dataview/histogram.js
```diff
@@ -97,16 +97,102 @@ var histogramQueryTpl = dot.template([
     'ORDER BY bin'
 ].join('\n'));
 
+var dateBasicsQueryTpl = dot.template([
+    'basics AS (',
```
ethervoid: We've to take into account that if the user's table is using `timestamp without timezone` this query is going to fail because it expects `timestampz`:

```
ERROR:  42883: function width_bucket(timestamp without time zone, timestamp with time zone[]) does not exist
LINE 39:         WIDTH_BUCKET(postedtime AT TIME ZONE '2', bins_array...
```

maybe we should cast the input to `timestampz` just to be completely sure?

lib/cartodb/backends/dataview.js
```diff
@@ -30,6 +32,7 @@ DataviewBackend.prototype.getDataview = function (mapConfigProvider, user, param
             mapConfigProvider.getMapConfig(this);
         },
         function runDataviewQuery(err, mapConfig) {
+            /* jshint maxcomplexity: 7 */
```
rochoa: If possible, I would try to avoid raising the max complexity by refactoring a bit the function.

lib/cartodb/backends/dataview.js
```diff
@@ -88,6 +91,10 @@ DataviewBackend.prototype.getDataview = function (mapConfigProvider, user, param
                 {ownFilter: ownFilter}
             );
 
+            if (params.aggregation && DATE_AGGREGATIONS.indexOf(params.aggregation) !== -1) {
```
rochoa: I usually tend to use a _set/dict_ for checking the existence. Not a big problem, but I prefer the verbs/methods you can use: `VALID_DATE_AGGREGATIONS.hasOwnProperty(aggregation)`.

lib/cartodb/controllers/base.js
```diff
@@ -25,6 +25,8 @@ var REQUEST_QUERY_PARAMS_WHITELIST = [
     'start', // number
     'end', // number
     'column_type', // string
+    'aggregation', //string
```
rochoa: What about taking the time to, finally, migrate the dataviews endpoints to use the [allow-query-params](https://github.com/CartoDB/Windshaft-cartodb/blob/02bf1dd2d7dee13071b815a8ba21466f22a81680/lib/cartodb/middleware/allow-query-params.js)
 as we already do in, for instance, [named-maps controller](https://github.com/CartoDB/Windshaft-cartodb/blob/02bf1dd2d7dee13071b815a8ba21466f22a81680/lib/cartodb/controllers/named_maps.js#L35).

lib/cartodb/models/dataview/histogram.js
rochoa: Good catch! Furthermore, we already defined a maximum number of bins limit of 48 because the frontend cannot render much more as the width of widget constraints the number of pixels you can render. This case is a bit different because the timeseries widget has different constraints, but the performance of the frontend side might suffer as well, so that's also an important consideration.

Another thing to have in mind is that the response format allows you to have a big number of bins without the need to send all of them but just the ones with actual data in them. So to generate a 5 million bins response we should have at least a 5 million dataset with each row falling behind each of the 5M bins. Something that might happen in the near future when we support bigger datasets.

lib/cartodb/backends/dataview.js
rochoa: BTW, I opened https://github.com/CartoDB/Windshaft-cartodb/pull/696, it would be nice if you can review and re-use it before merging this one 🙏 .

lib/cartodb/models/dataview/histogram.js
jorgesancha: The limit on the number of buckets seems to be 256, no?
image

Or am I missing something here?

rochoa: Sorry for misleading you, I was talking about histograms in general and, especially, when the number of bins is not defined by the user. In that case, we use https://en.wikipedia.org/wiki/Freedman%E2%80%93Diaconis_rule to select the number of bins, but up to 48 to avoid issues on the frontend side.

As I was saying, the constraints for time-series are a bit different in the scale, but we should have the same kind of problems. For instance, it doesn't make sense to use 1000 buckets if you have to render them in 300px@2x.

dgaubert: We cannot use both `aggregation` and `bins` (aka buckets) at the same time because it isn't possible to define the numbers of bins and its width for the same histogram. In order to create a histogram we have two paths:
  1) Defining the number of buckets (bins) for numeric and date columns (current histogram)
  2) Defining the aggregation (width of each bin) for date columns (new histogram)

dgaubert: ⬆️ might be important to @ivanmalagon

lib/cartodb/controllers/base.js
dgaubert: I forgot we had this.

Done.

lib/cartodb/backends/dataview.js
dgaubert: Done.

lib/cartodb/models/dataview/histogram.js
dgaubert: Yep, @ivanmalagon has just fixed that! Thanks mates! :)

dgaubert:
I'd say it's ready for another round of CR. Any brave coder around?

lib/cartodb/models/dataview/histogram.js
```diff
@@ -10,6 +10,47 @@ var columnTypeQueryTpl = dot.template(
...
+    'WITH',
+    'dates AS (',
+    '    SELECT',
+    '        MAX({{=it.column}}) AS _end,',
+    '        MIN({{=it.column}}) AS _start',
+    '    FROM ({{=it.query}}) _cdb_source',
+    '),',
+    'interval_in_days AS (',
+    '    SELECT' ,
+    '        DATE_PART(\'day\', _end - _start) AS days',
```
jgoizueta: It seems that this works only with `timestamp`s, not `date` or `time` columns, so we should cast the user column accordingly.

dgaubert: Fixed! Thanks for tip!

lib/cartodb/models/dataview/histogram.js
```diff
@@ -163,9 +300,12 @@ Histogram.prototype.sql = function(psql, override, callback) {
     }
 
     if (this._columnType === 'date') {
-        _column = columnCastTpl({column: _column});
+        return this._buildDateHistogramQuery(psql, override, callback);
     }
 
+    if (this._columnType === 'date') {
+        _column = columnCastTpl({ column: this.column});
```
jgoizueta: This cannot be reached now, can it?

dgaubert:
@rochoa I've just uploaded support for `aggregation: 'auto'` keeping previous behaviour to not break the current interfaz for histograms.

It's ready for CR.

PD: also it's synced with `master`

rochoa:
Sorry, I had some comments but totally forgot to "start the review". I just did.

lib/cartodb/models/dataview/histogram.js
```diff
@@ -184,16 +325,18 @@ Histogram.prototype.sql = function(psql, override, callback) {
...
+    this._buildQuery(psql, override, callback);
 };
 
-Histogram.prototype._buildQuery = function (override) {
+Histogram.prototype._buildQuery = function (psql, override, callback) {
     var filteredQuery, basicsQuery, binsQuery;
     var _column = this.column;
     var _query = this.query;
 
+    if (this._columnType === 'date' && this.aggregation !== undefined) {
```
dgaubert: Totally agree. In fact, that was my first approach 😄  but adding a new type of histogram exposed in the API (`{ type: "date-histogram", options: { ... } }`). This code smells like a`state/strategy` pattern a lot. Also I want to see this, first, working in production and then refactor it.

lib/cartodb/backends/dataview.js
```diff
@@ -104,6 +114,12 @@ function getOverrideParams(params, ownFilter) {
         },
         {ownFilter: ownFilter}
     );
+
+    if (params.aggregation && DATE_AGGREGATIONS.hasOwnProperty(params.aggregation)) {
+        overrideParams.aggregation = params.aggregation;
```
dgaubert: We should throw an error definitely.

lib/cartodb/models/dataview/histogram.js
rochoa: No, I'm not talking that much about the public API, but more about how to improve the implementation.

dgaubert: Sure, I got it. 😃

rochoa:
Quick question: how should I run the tests? Do I need to run them with a specific timezone setting?

dgaubert:
Nope, tests do not rely on any environment timezone setting. Did you found something weird?

rochoa:
Yes. 😕 

```
$ ./run_tests.sh test/acceptance/dataviews/histogram.js --grep quarter
[...]
Running tests


  histogram-dataview for date column type
    1) should aggregate histogram using "quarter" aggregation supporting timestamp with offset
    2) should aggregate histogram using "quarter" aggregation supporting timestamp without offset


  0 passing (451ms)
  2 failing

  1) histogram-dataview for date column type should aggregate histogram using "quarter" aggregation supporting timestamp with offset:

      Uncaught AssertionError: '2006-12-31T23:00:00Z' == '2007-01-01T00:00:00Z'
      + expected - actual

      -2006-12-31T23:00:00Z
      +2007-01-01T00:00:00Z

      at test/acceptance/dataviews/histogram.js:457:24
      at Function.finish (test/support/test-client.js:409:20)
      at next (node_modules/step/lib/step.js:51:23)
      at test/support/test-client.js:402:21
      at Server.<anonymous> (test/support/assert.js:169:17)
      at emitCloseNT (net.js:1554:8)
      at _combinedTickCallback (internal/process/next_tick.js:77:11)
      at process._tickCallback (internal/process/next_tick.js:104:9)

  2) histogram-dataview for date column type should aggregate histogram using "quarter" aggregation supporting timestamp without offset:

      Uncaught AssertionError: '2006-12-31T23:00:00Z' == '2007-01-01T00:00:00Z'
      + expected - actual

      -2006-12-31T23:00:00Z
      +2007-01-01T00:00:00Z

      at test/acceptance/dataviews/histogram.js:457:24
      at Function.finish (test/support/test-client.js:409:20)
      at next (node_modules/step/lib/step.js:51:23)
      at test/support/test-client.js:402:21
      at Server.<anonymous> (test/support/assert.js:169:17)
      at emitCloseNT (net.js:1554:8)
      at _combinedTickCallback (internal/process/next_tick.js:77:11)
      at process._tickCallback (internal/process/next_tick.js:104:9)
```

dgaubert:
🤔 First time I see that error neither in my local env nor `travis-ci`.

dgaubert:
Oh wait, please check that your postgres server has UTC as default timezone in `postgresql.conf` file

rochoa:
My local env:

```
$ date
Wed Aug  2 15:50:31 CEST 2017
$ TZ=UTC date
Wed Aug  2 13:50:36 UTC 2017
```

But with:

```
$ TZ=UTC ./run_tests.sh test/acceptance/dataviews/histogram.js --grep quarter
```

Tests also fail, with same results.

dgaubert:
I didn't know `TZ=UTC`. Pretty cool BTW. @rubenmoya had a very same issue and we fixed modifying `postgresql.conf` and restarting the server:

```
SET TIME ZONE 'UTC'
```

rochoa:
Oh! Thank you. I had my local PostgreSQL instance setup with timezone=Europe/Madrid. After your comment, I changed it to Etc/UTC and the tests are working as expected.

lib/cartodb/models/dataview/histogram.js
```diff
@@ -434,12 +434,29 @@ Histogram.prototype._shouldOverrideBins = function (override) {
...
+    'year': true
+};
+
 Histogram.prototype._buildDateHistogramQuery = function (psql, override, callback) {
     var _column = this.column;
     var _query = this.query;
     var _aggregation = override && override.aggregation ? override.aggregation : this.aggregation;
     var _offset = override && Number.isFinite(override.offset) ? override.offset : this.offset;
 
+    if (!DATE_AGGREGATIONS.hasOwnProperty(_aggregation)) {
```
rochoa: Much better! 👍

dgaubert: 🇪🇸 ????

PR #700, opened 2017-06-08
Handle Infinity and NaN values for dataviews

dgaubert:
Fixes https://github.com/CartoDB/cartodb/issues/12154
 - [x] Histogram
 - [x] Histogram (Overview)
 - [x] Formula
 - [x] Formula (Overview)
 - [x] Aggregation
 - [x] Aggregation (Overview)


lib/cartodb/models/dataview/aggregation.js
```diff
@@ -84,7 +108,7 @@ var TYPE = 'aggregation';
      }
  }
  */
-function Aggregation(query, options) {
+function Aggregation(query, options, queries) {
```
jgoizueta: what's the purpose of this `queries` parameter?

dgaubert: it's an object that contains, among others, the original query w/o filters. We use that sql to figure out the `pg_type` of the column, if the filter does not return data then it's not possible to know what type of column we are handling.

lib/cartodb/models/dataview/base.js
```diff
@@ -24,3 +27,41 @@ BaseDataview.prototype.getResult = function(psql, override, callback) {
 BaseDataview.prototype.search = function(psql, userQuery, callback) {
     return callback(null, this.format({ rows: [] }));
 };
+
+var FLOAT_OIDS = {
+    700: true,
+    701: true
+};
```
jgoizueta: It seems that type `Numeric`(1700) also has 'NaN' values, but no infinities.
Since infinity detection is being done here by comparing values float infinities (`'Infinity'::float`, etc) I think that just by adding the type 1700 here we would gain NaN support for Numeric columns. I suggest doing so and adding tests that use `'NaN'::Numeric` values.

dgaubert:
Hey @jgoizueta, I've just pushed a couple of commits to support numeric NaN values for dataviews

jgoizueta:
Great! I think this is ready for acceptance.

PR #710, opened 2017-07-14
replace timezone => offset

matallo:
- replaced argument nomenclature `timezone` => `offset` so I can work with a different `timezone` as a string and still set an `offset` at Builder, otherwise it would be prone to confusion


dgaubert:
If tests are green, go ahead

matallo:
i'm solving the conflicts, merging as the tests are green!

PR #738, opened 2017-09-01
2.43.1 fix pg vulnerability

oleurud:
Adding a new cartodb-psql fixing a pg vulnerability ("cartodb-psql": "0.10.1")

This will be the tag 2.43.2

-----------------

More info about pg vulnerability 
https://node-postgres.com/announcements#2017-08-12-code-execution-vulnerability

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,25 @@
 dist: precise
 
-before_install:		
-  - lsb_release -a
-  - sudo mv /etc/apt/sources.list.d/pgdg.list* /tmp
-  - sudo apt-get -qq purge postgis* postgresql*
-  - sudo rm -Rf /var/lib/postgresql /etc/postgresql
-  - sudo apt-add-repository --yes ppa:cartodb/postgresql-9.3
-  - sudo apt-add-repository --yes ppa:cartodb/gis
-  - sudo apt-get update
-  - sudo apt-get install pkg-config libgif-dev libjpeg-dev libcairo2-dev libpango1.0-dev
-  - sudo apt-get install -q postgresql-9.3-postgis-2.1
-  - sudo apt-get install -q postgresql-contrib-9.3
-  - sudo apt-get install -q postgresql-plpython-9.3
-  - sudo apt-get install -q postgis
-  - sudo apt-get install -q gdal-bin
-  - sudo apt-get install -q ogr2ogr2-static-bin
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,19 +1,27 @@
...
 
-addons:
-  postgresql: "9.3"
-  apt:
-    packages:
-      - pkg-config
-      - libcairo2-dev
-      - libjpeg8-dev
-      - libgif-dev
+before_install:
```
dgaubert: I'd try to follow [this](https://github.com/CartoDB/Windshaft-cartodb/commit/28a2c29a399fc96a73417e346f2a53005511e454) with `postgresql-9.3-postgis-2.1` and avoid to use `sudo`.

.travis.yml
```diff
@@ -1,19 +1,27 @@
...
+  - sudo apt-get -qq purge postgis* postgresql*
+  - sudo rm -Rf /var/lib/postgresql /etc/postgresql
+  - sudo apt-add-repository --yes ppa:cartodb/postgresql-9.3
+  - sudo apt-add-repository --yes ppa:cartodb/gis
+  - sudo apt-get update
+  - sudo apt-get install pkg-config libgif-dev libjpeg-dev libcairo2-dev libpango1.0-dev
+  - sudo apt-get install -q postgresql-9.3-postgis-2.1
+  - sudo apt-get install -q postgresql-contrib-9.3
+  - sudo apt-get install -q postgresql-plpython-9.3
+  - sudo apt-get install -q postgis
```
dgaubert: here it's installing `postgis` twice. isn't it?

.travis.yml
oleurud: I am going to try it

.travis.yml
oleurud: I am not sure about it. We do it in the same way in CartoDB-SQL-API https://github.com/CartoDB/CartoDB-SQL-API/blob/master/.travis.yml and I did the same without much thought

oleurud: Looking into https://github.com/CartoDB/Windshaft-cartodb/blob/master/.travis.yml we only install       postgresql-9.5-postgis-2.3 and not postgis, probably you are right

.travis.yml
oleurud: Done! works fine

oleurud:
The tag 2.43.2 was created

PR #742, opened 2017-09-06
Fix out of range bug in date histograms

ivanmalagon:
Fix for the bug reported at https://github.com/CartoDB/support/issues/962

The problem was that the total range between max and min timestamps is above the range of `integer` but both operands are within `integer` limits. PostgreSQL try to accommodate the result and it doesn't fit so an `out of range` error is triggered.

Making both values as double precision (float) fixes the bug.

This is consistent with the basics query with no overrides, since it gets the values from date using `datepart` function, that returns double precision.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 3.12.8
+Released 2017-mm-dd
+
+Bug fixes:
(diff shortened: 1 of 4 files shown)


dgaubert:
Nice catch @ivanmalagon

Thank you so much for tackling this issue.

Just a minor detail, could you update NEWS with info relevant to this PR.

After that, please feel free to merge it.

PR #744, opened 2017-09-11
Histogram refactor

dgaubert:
Big refactor in Histogram dataview:
  - Split histogram dataview in two implementations: NumericHistogram and DateHistogram.
  - Applied state pattern.
  - Created a BaseHistogram to reuse common code between both implementations.
  - Lots of ES6's syntax sugar goodies applied.
  - Template method pattern applied in `.format` method.
 
cc/ @oleurud


lib/cartodb/models/dataview/histograms/base-histogram.js
```diff
@@ -0,0 +1,85 @@
...
+
+        this.query = query;
+        this.queries = queries;
+        this.column = options.column;
+        this.bins = options.bins;
+
+        this._columnType = null;
+    }
+
+    sql (psql, override, callback) {
```
enekid: Optional parameters before a mandatory callback look a little weird to me.

What do you think about using [named parameters via destructuring](http://exploringjs.com/es6/ch_parameter-handling.html#_named-parameters-via-destructuring)?
```javascript
sql ({ sql, override = {} }, callback) {

  if (this._columnType === null) { ...
```

Calls to this method would be like this:
```javascript
const sql = 'SQL';
const override = { a: 1, b: 2 };

baseHistogram.sql({ sql: 'SQL' }, callback);
baseHistogram.sql({ sql }, callback);
baseHistogram.sql({ sql, override }, callback);
```

[stackoverflow](https://stackoverflow.com/questions/12826977/multiple-arguments-vs-options-object)

dgaubert: Yeah good catch!

With ES6 we can mitigate some of the gotchas of node callback pattern. Giving a try to see how it looks ...

dgaubert: But.... there is a problem... the method `.sql()` is part of the public API shared among all dataviews and it's called by BaseDataview while fetching the result of any of them. That means that we have to change the signature on every dataview's implementation (category, formula, etc..)

I like your point of view, but it's out of the scope of this refactor, feel free to create a new PR applying the changes that you are proposing.

PR #747, opened 2017-09-15
Formula dataview refactor

dgaubert:
Lots of ES6's syntax sugar goodies applied:
  - Template strings to build formula query.
  - New class syntax
  - Renamed base class
  - Remove every reference to widgets.

### DIFF
--- a/lib/cartodb/models/dataview/formula.js
+++ b/lib/cartodb/models/dataview/formula.js
@@ -53,6 +53,16 @@ module.exports = class Formula extends BaseDataview {
     constructor (query, options = {}, queries = {}) {
         super();
 
+        this._checkOptions(options);
+
+        this.query = query;
+        this.queries = queries;
+        this.column = options.column || '1';
+        this.operation = options.operation;
+        this._isFloatColumn = null;
+    }
+
+    _checkOptions (options) {
         if (typeof options.operation !== 'string') {
             throw new Error(`Formula expects 'operation' in dataview options`);
         }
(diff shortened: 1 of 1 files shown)


test/acceptance/dataviews/overviews.js
```diff
@@ -88,7 +102,14 @@ describe('dataviews using tables without overviews', function() {
                     if (err) {
                         return done(err);
                     }
-                    assert.deepEqual(formula_result, { operation: 'count', result: 256, nulls: 0, type: 'formula' });
+                    assert.deepEqual(formula_result, {
+                        operation: 'count',
+                        result: 256,
+                        nulls: 0,
+                        infinities: 0,
```
rochoa: Why do we need to change the tests this being a refactor?

dgaubert: Actually they are bad tests. All of them was missing special float values. As we do in other dataviews. So I fixed them.

rochoa: It feels weird to me, as only the expectation is changing. In these cases, I usually prefer to do two PR: one with the tests changed the second one with the refactor.

dgaubert: You are right, my mind was in rush when I figured out what was going on here.

lib/cartodb/models/dataview/formula.js
```diff
@@ -1,36 +1,44 @@
...
-    '  {{=it._column}} != \'-infinity\'::float',
-    'AND',
-    '  {{=it._column}} != \'NaN\'::float{{?}}'
-].join('\n'));
-
-var VALID_OPERATIONS = {
+const BaseDataview = require('./base');
+const debug = require('debug')('windshaft:dataview:formula');
+
+const countInfinitiesQueryTpl = ctx => `
```
oleurud: :rocket:

lib/cartodb/models/dataview/formula.js
```diff
@@ -1,36 +1,44 @@
...
+    SELECT count(1) FROM (${ctx.query}) __cdb_formula_infinities
+    WHERE ${ctx.column} = 'infinity'::float OR ${ctx.column} = '-infinity'::float
+`;
+
+const countNansQueryTpl = ctx => `
```
oleurud: :rocket:

PR #746, opened 2017-09-14
List dataview refactor

dgaubert:
Lots of ES6's syntax sugar goodies applied:
  - Template strings to build aggregation query.
  - New class syntax
  - Renamed base class
  - Remove every reference to widgets.

PR #748, opened 2017-09-15
Base dataview refactor

dgaubert:
This PR depends on #744, #745, #746 and #747. Using ES6 class syntax  brings some checks while consuming it. For instance, aggregation's constructor calls to `BaseDataview.apply(this)` in oder to perform ES5 classic inheritance raising the following error: `Class constructor BaseDataview cannot be invoked without 'new'`


lib/cartodb/models/dataview/base.js
```diff
@@ -1,71 +1,60 @@
...
 function getPGTypeName (pgType) {
     return {
         float: FLOAT_OIDS.hasOwnProperty(pgType),
         date: DATE_OIDS.hasOwnProperty(pgType)
     };
 }
+
+module.exports = class BaseDataview {
+    getResult (psql, override, callback) {
+        const self = this;
```
rochoa: Why not use an arrow function so you keep the context for `this`?

dgaubert: Cool! 

On it

PR #749, opened 2017-09-15
Dataview factory refactor

dgaubert:
Uses ES6 class syntax to define static properties and methods.

`<rant>`
After using ES6 class syntax during some time I've realised that in Node we expose modules in many different ways. In some modules we expose a simple function or a collection of functions or an object with a bunch of methods or a classical function that acts like a class constructor (ES5). That's cool and you can expose whatever you want but in projects with a large code base I find that a little confusing, because you never know what it's being exposed when you require a module and you end reading the code of the required module in order to know how to use it.

In conclusion, IMHO exposing always a class you can do whatever you need to expose and it's pretty predictable for other developers.

Thoughts?
`</rant>`

### DIFF
--- a/lib/cartodb/models/dataview/factory.js
+++ b/lib/cartodb/models/dataview/factory.js
@@ -9,10 +9,12 @@ module.exports = class DataviewFactory {
     }
 
     static getDataview (query, dataviewDefinition) {
-        const type = dataviewDefinition.type;
+        const { type, options, sql } = dataviewDefinition;
+
         if (!this.dataviews[type]) {
             throw new Error('Invalid dataview type: "' + type + '"');
         }
-        return new this.dataviews[type](query, dataviewDefinition.options, dataviewDefinition.sql);
+
+        return new this.dataviews[type](query, options, sql);
     }
 };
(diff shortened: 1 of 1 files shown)


lib/cartodb/models/dataview/factory.js
```diff
@@ -1,18 +1,20 @@
...
+module.exports = class DataviewFactory {
+    static get dataviews() {
+        return Object.keys(dataviews).reduce((allDataviews, dataviewClassName) => {
+            allDataviews[dataviewClassName.toLowerCase()] = dataviews[dataviewClassName];
+            return allDataviews;
+        }, {});
+    }
+
+    static getDataview (query, dataviewDefinition) {
+        const { type, options, sql } = dataviewDefinition;
```
oleurud: :rocket:

oleurud:
About your question, I am not sure about what is the better way to expose code from a module, IMHO is hard to select only a unique way to expose code.

A class, for a medium module, is nice for me. Only add to your explanation that maybe the class needs some "private" methods, and you will expose it (and probably, you don't want it). 

For big or medium size modules, maybe you need to expose more than one class for example. And for small or medium modules, maybe a small function is a pretty solution.

PR #751, opened 2017-09-21
Middlewarify

dgaubert:


rochoa:
I already opened a PR (https://github.com/CartoDB/Windshaft-cartodb/pull/721) to move the LZMA handling to a middleware.

Ideally, that should only be configured in endpoints where the `config` param is used. I don't care much about what you prefer to do with that PR, but whatever you decide, take care of closing it.

lib/cartodb/controllers/named_maps.js
```diff
@@ -9,6 +9,7 @@ var BaseController = require('./base');
 var cors = require('../middleware/cors');
```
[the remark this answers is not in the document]
dgaubert: I like the idea, but I'd wait to see the final folder structure, we are going to create lots of middleware and we don't know where they are going to be placed yet.

lib/cartodb/controllers/layergroup.js
```diff
@@ -43,39 +45,57 @@ function LayergroupController(authApi, pgConnection, mapStore, tileBackend, prev
...
+    this.prepareContext = prepareContextMiddleware(authApi, pgConnection);
 }
 
 util.inherits(LayergroupController, BaseController);
 
 module.exports = LayergroupController;
 
+LayergroupController.prototype.register = function(router) {
+    router.use(
+        cors(),
```
enekid: There is an official express.js [cors middleware](https://www.npmjs.com/package/cors). Should we use it?

We also should consider using [helmet](https://www.npmjs.com/package/helmet)

oleurud:
After some tries, finally, I have committed an ugly solution for a conflicting route case.

The problem:
We have a conflict between this routes and sometimes we have unexpected behavior:
- '/:token/:layer/:z/:x/:y.(:format)' (layergroup.js)
- '/static/named/:template_id/:width/:height.:format' (named_maps.js)

I have tried to resolve the problem matching the "token" param in the correct case. I have shown the token has this pattern: {user}@{tpl_id}@{realToken}:{cache_buster} but only the realToken is required.

So, we have at least this cases (the are more, this cases are reflected in tests) to match the token param:
'/:user@:tpl@:token::cache_buster',
'/:user@:tpl@:token',
'/:token::cache_buster',
'/:token/:layer/:z/:x/:y.(:format)',

and the last option is a big problem, because is basically the same as '/static/named/:template_id/:width/:height.:format'

At this point, we have 2 options: 
- the ugly solution of the middleware (I have committed this)
- use a regexp check: token !== 'static'

I have not implemented the second because I think it would be necessary to create a middleware to create the token param in req (req.params.token) and it would be an ugly solution too.

Any idea?

lib/cartodb/middleware/context/authorize.js
```diff
@@ -0,0 +1,27 @@
+const _ = require('underscore');
+
+module.exports = function authorizeMiddleware (authApi) {
+    return function (req, res, next) {
+        // FIXME: Temporary hack to share data between middlewares. Express overrides req.params to
+        // parse url params to an object and it's performed after matching path and controller.
+        req.locals = {};
```
enekid: I think it would be more future-proof to stick to the documented way: ```res.locals``` instead of ```req.locals```.

[Stackoverflow](https://stackoverflow.com/a/33451885)
[Express Api Documentation](http://expressjs.com/en/api.html#res.locals)

oleurud: You are right, I am going to change it

lib/cartodb/controllers/layergroup.js
dgaubert: For time being, let's keep our module to not add more entropy to this PR. But please, open a PR with your proposal. 😈

lib/cartodb/middleware/context/authorize.js
dgaubert: Nice catch, that's one of the first things I did in this PR to pass info to error middleware. I didn't have `res` param because `req2param` just received `req` so I had to use `req.locals`.

Thanks.

lib/cartodb/controllers/layergroup.js
enekid: Ok! I'll wait until all this code refactor settles a bit.

oleurud: Another interesting package is https://github.com/nodesecurity/nsp to check vulnerabilities in dependencies

dgaubert: Please, could you open an issue with your proposals?

PR #753, opened 2017-09-22
Layergroup Token parsing as middleware

rochoa:
Reuses LayergroupToken model from tests.

### DIFF
--- a/lib/cartodb/middleware/layergroup-token.js
+++ b/lib/cartodb/middleware/layergroup-token.js
@@ -1,0 +1,35 @@
+var LayergroupToken = require('../models/layergroup-token');
+
+module.exports = function layergroupTokenMiddleware(req, res, next) {
+    if (!req.params.hasOwnProperty('token')) {
+        return next();
+    }
+
+    var user = req.context.user;
+
+    var layergroupToken = LayergroupToken.parse(req.params.token);
+    req.params.token = layergroupToken.token;
+    req.params.cache_buster = layergroupToken.cacheBuster;
+
+    if (layergroupToken.signer) {
+        req.params.signer = layergroupToken.signer;
+        if (!req.params.signer) {
+            req.params.signer = user;
+        } else if (req.params.signer !== user) {
+            var statusCode = 403;
+            if (req.query && req.query.callback) {
+                statusCode = 200;
+            }
+            var errorMessage = `Cannot use map signature of user "${req.params.signer}" on db of user "{${user}"`;
+            return res.status(statusCode).json({
+                errors: [errorMessage],
+                errors_with_context: [{
+                    type: 'auth',
+                    message: errorMessage
+                }]
+            });
+        }
+    }
+
+    return next();
+};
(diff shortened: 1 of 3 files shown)


lib/cartodb/middleware/layergroup-token.js
```diff
@@ -0,0 +1,35 @@
...
+        req.params.signer = layergroupToken.signer;
+        if (!req.params.signer) {
+            req.params.signer = user;
+        } else if (req.params.signer !== user) {
+            var statusCode = 403;
+            if (req.query && req.query.callback) {
+                statusCode = 200;
+            }
+            var errorMessage = `Cannot use map signature of user "${req.params.signer}" on db of user "{${user}"`;
+            return res.status(statusCode).json({
```
rochoa: This is ugly as hell. But I had this code in a stash for a while, and now that you are working on the middleware thingy, I thought it was nice to share so you can decide what to do with this code.

dgaubert: Haha!

we are duplicating some functionality to create `error` and `errors_with_context` response. It'd be great to yield the control to error-middleware:
```
const err = new Error(message);
err.http_status = statusCode;
err.type = 'auth'
return next(err);
```

test/acceptance/ported/support/ported_server_options.js
```diff
@@ -56,9 +55,12 @@ module.exports = _.extend({}, serverOptions, {
 
         // this is in case you want to test sql parameters eg ...png?sql=select * from my_table limit 10
         req.params =  _.extend({}, req.params);
-        if (req.params.token) {
```
dgaubert: Damm! Ported tests!!

:man_shrugging:

lib/cartodb/middleware/layergroup-token.js
rochoa: Yes! I didn't know much about that, but I knew you had something to avoid this duplication. 💪

dgaubert: The underlaying idea is to be able to categorise every and then, depending on content-negotiation, respond with json, jsonp, text or whatever in error-middleware

dgaubert: BTW koa provides helpers to deal with `content-negotiation`.  From docs:

```
# request.is(types...)

Check if the incoming request contains the "Content-Type" header field, and it contains any of the give mime types. If there is no request body, null is returned. If there is no content type, or the match fails false is returned. Otherwise, it returns the matching content-type.

// With Content-Type: text/html; charset=utf-8
ctx.is('html'); // => 'html'
ctx.is('text/html'); // => 'text/html'
ctx.is('text/*', 'text/html'); // => 'text/html'

// When Content-Type is application/json
ctx.is('json', 'urlencoded'); // => 'json'
ctx.is('application/json'); // => 'application/json'
ctx.is('html', 'application/*'); // => 'application/json'

ctx.is('html'); // => false
```

Checking if `express` has something similar... 😈

dgaubert: Yep, same interface indeed!

rochoa: It does have: https://github.com/CartoDB/Windshaft-cartodb/blob/33121871b0df4d90f8d6712301e6509f57b5618e/lib/cartodb/controllers/map.js#L87

PR #762, opened 2017-09-28
Add configuration flag to enable/disable direct PgSQL MVT

dv343:
Needed to use the new functionality of Windshaft: https://github.com/CartoDB/Windshaft/pull/561


config/environments/development.js.example
```diff
@@ -107,6 +107,9 @@ var config = {
       // Milliseconds since last access before renderer cache item expires
       cache_ttl: 60000,
       statsInterval: 5000, // milliseconds between each report to statsd about number of renderers and mapnik pool status
+      mvt: {
+        usePgSQL: false
```
rochoa: This must be documented: how it works, what are the requirements, and so.

And, what about using something closer to the real thing: `usePostgisAsMVT`?

dv343: I'm gonna document it, but I wouldn't use that name, since we already are inside the `mvt` conf, also, this way is more "flexible", this PR is basically an implementation leak, but we don't need to leak even the PostGIS function names.

rochoa: Then, what `usePgSQL` refers to?

dv343:
I've seen that your comment disappeared... `usePgSQL` refers to using PostgreSQL directly, but it doesn't says how.

config/environments/development.js.example
rafatower: `usePostGisMvtRenderer`

my 2 cents

test/acceptance/mvt.js
```diff
@@ -48,16 +140,365 @@ describe('mvt', function () {
...
+                    cartodb_id: 1109,
+                    pop_max: 71373,
+                    name: "Mardin",
+                    adm0name: "Turkey"
+                });
+                done();
+            });
+        });
+
+        it.skip('should work with mapnik substitution tokens', function (done) {
```
rochoa: Why do we skip this test?

dv343: Because it triggers the MVT v2 compliance problem

rochoa: Which one exactly? Anyway, we should skip it only if the Postgis mode is enabled, right?

dv343: This one, this is the error trace:
```  3) mvt (postgis) use only needed columns should work with mapnik substitution tokens:
     Uncaught TypeError: Vector Tile has POLYGON with first ring clockwise. It is not valid according to v2 of VT spec.
      at TypeError (native)
      at test/acceptance/mvt.js:372:50
      at Function.finish (test/support/test-client.js:743:20)
      at next (node_modules/step/lib/step.js:51:23)
      at Server.<anonymous> (test/support/assert.js:124:24)
      at emitCloseNT (net.js:1556:8)
      at _combinedTickCallback (internal/process/next_tick.js:77:11)
      at process._tickCallback (internal/process/next_tick.js:104:9)

```

dv343: "only if the Postgis mode is enabled, right?" Well... yes

rochoa: I see. However, I think we need to keep a test for _should work with mapnik substitution tokens_, by removing the `Vector Tile has POLYGON with first ring clockwise` problem (making the query not to generate an invalid Vector Tile polygon), but also create a new test to keep track of the problem with `Vector Tile has POLYGON with first ring clockwise`.

dv343: I think it's safe to be ignored for now, we are testing similar functionality with other tests: https://github.com/CartoDB/Windshaft-cartodb/search?utf8=%E2%9C%93&q=%22substitution+tokens%22&type=

Removing the validity problem can be difficult.

rochoa: Where are we testing MVT + Substitution tokens?

test/acceptance/mvt.js
```diff
@@ -325,7 +325,8 @@ return function () {
             });
         });
 
-        it.skip('should work with mapnik substitution tokens', function (done) {
+        var skipOnPostGIS = usePostGIS ? it.skip: it;
```
rochoa: 👍

PR #765, opened 2017-09-29
req.params to res.locals

oleurud:
Not use "req.params" (only for URL parsed params) for our logic, and use "res.locals" instead of

Maintain the scope small to avoid changes in other libraries.

### DIFF
--- a/lib/cartodb/api/auth_api.js
+++ b/lib/cartodb/api/auth_api.js
@@ -105,7 +105,7 @@ AuthApi.prototype.authorize = function(req, res, callback) {
             }
 
             // authorized by api key, login as the given username and stop
-            self.pgConnection.setDBAuth(user, res.locals.db, function(err) {
+            self.pgConnection.setDBAuth(user, res.locals, function(err) {
                 callback(err, true); // authorized (or error)
             });
         },
@@ -128,7 +128,7 @@ AuthApi.prototype.authorize = function(req, res, callback) {
                 return callback(null, false);
             }
 
-            self.pgConnection.setDBAuth(user, res.locals.db, function(err) {
+            self.pgConnection.setDBAuth(user, res.locals, function(err) {
                 req.profiler.done('setDBAuth');
                 callback(err, true); // authorized (or error)
             });
--- a/lib/cartodb/api/auth_api.js
+++ b/lib/cartodb/api/auth_api.js
@@ -26,15 +26,15 @@ module.exports = AuthApi;
 //                 null if the request is not signed by anyone
 //                 or will be a string cartodb username otherwise.
 //
-AuthApi.prototype.authorizedBySigner = function(params, callback) {
-    if ( ! params.token || ! params.signer ) {
+AuthApi.prototype.authorizedBySigner = function(res, callback) {
+    if ( ! res.locals.token || ! res.locals.signer ) {
         return callback(null, false); // no signer requested
     }
 
     var self = this;
 
-    var layergroup_id = params.token;
-    var auth_token = params.auth_token;
+    var layergroup_id = res.locals.token;
+    var auth_token = res.locals.auth_token;
 
     this.mapStore.load(layergroup_id, function(err, mapConfig) {
         if (err) {
@@ -87,9 +87,9 @@ AuthApi.prototype.authorizedByAPIKey = function(user, req, callback) {
  * @param params res.locals parameters. Contains the auth parameters
  * @param callback function(err, allowed) is access allowed not?
  */
-AuthApi.prototype.authorize = function(req, params, callback) {
+AuthApi.prototype.authorize = function(req, res, callback) {
     var self = this;
-    var user = params.user;
+    var user = res.locals.user;
 
     step(
         function () {
@@ -102,11 +102,11 @@ AuthApi.prototype.authorize = function(req, params, callback) {
             // if not authorized by api_key, continue
             if (!authorized)  {
                 // not authorized by api_key, check if authorized by signer
-                return self.authorizedBySigner(params, this);
+                return self.authorizedBySigner(res, this);
             }
 
             // authorized by api key, login as the given username and stop
-            self.pgConnection.setDBAuth(user, params, function(err) {
+            self.pgConnection.setDBAuth(user, res.locals, function(err) {
                 callback(err, true); // authorized (or error)
             });
         },
@@ -121,7 +121,7 @@ AuthApi.prototype.authorize = function(req, params, callback) {
                 // if no signer name was given, let dbparams and
                 // PostgreSQL do the rest.
                 //
-                if ( ! params.signer ) {
+                if ( ! res.locals.signer ) {
                     return callback(null, true); // authorized so far
                 }
 
@@ -129,7 +129,7 @@ AuthApi.prototype.authorize = function(req, params, callback) {
                 return callback(null, false);
             }
 
-            self.pgConnection.setDBAuth(user, params, function(err) {
+            self.pgConnection.setDBAuth(user, res.locals, function(err) {
                 req.profiler.done('setDBAuth');
                 callback(err, true); // authorized (or error)
             });
--- a/lib/cartodb/api/auth_api.js
+++ b/lib/cartodb/api/auth_api.js
@@ -86,7 +86,7 @@ AuthApi.prototype.authorizedByAPIKey = function(user, req, callback) {
  * @param req - standard req object. Importantly contains table and host information
  * @param callback function(err, allowed) is access allowed not?
  */
-AuthApi.prototype.authorize = function(req, res, callback) {
+AuthApi.prototype.authorize = function(req, params, callback) {
     var self = this;
     var user = req.context.user;
 
@@ -101,11 +101,11 @@ AuthApi.prototype.authorize = function(req, res, callback) {
             // if not authorized by api_key, continue
             if (!authorized)  {
                 // not authorized by api_key, check if authorized by signer
-                return self.authorizedBySigner(res.locals, this);
+                return self.authorizedBySigner(params, this);
             }
 
             // authorized by api key, login as the given username and stop
-            self.pgConnection.setDBAuth(user, res.locals, function(err) {
+            self.pgConnection.setDBAuth(user, params, function(err) {
                 callback(err, true); // authorized (or error)
             });
         },
@@ -120,7 +120,7 @@ AuthApi.prototype.authorize = function(req, res, callback) {
                 // if no signer name was given, let dbparams and
                 // PostgreSQL do the rest.
                 //
-                if ( ! res.locals.signer ) {
+                if ( ! params.signer ) {
                     return callback(null, true); // authorized so far
                 }
 
@@ -128,7 +128,7 @@ AuthApi.prototype.authorize = function(req, res, callback) {
                 return callback(null, false);
             }
 
-            self.pgConnection.setDBAuth(user, res.locals, function(err) {
+            self.pgConnection.setDBAuth(user, params, function(err) {
                 req.profiler.done('setDBAuth');
                 callback(err, true); // authorized (or error)
             });
--- a/lib/cartodb/api/auth_api.js
+++ b/lib/cartodb/api/auth_api.js
@@ -84,7 +84,7 @@ AuthApi.prototype.authorizedByAPIKey = function(user, req, callback) {
  * Check access authorization
  *
  * @param req - standard req object. Importantly contains table and host information
- * @param params res.locals parameters. Contains the auth parameters
+ * @param res - standard res object. Contains the auth parameters in locals
  * @param callback function(err, allowed) is access allowed not?
  */
 AuthApi.prototype.authorize = function(req, res, callback) {
--- a/lib/cartodb/api/auth_api.js
+++ b/lib/cartodb/api/auth_api.js
@@ -88,7 +88,7 @@ AuthApi.prototype.authorizedByAPIKey = function(user, req, callback) {
  */
 AuthApi.prototype.authorize = function(req, params, callback) {
     var self = this;
-    var user = req.context.user;
+    var user = params.user;
 
     step(
         function () {
--- a/lib/cartodb/api/auth_api.js
+++ b/lib/cartodb/api/auth_api.js
@@ -26,15 +26,15 @@ module.exports = AuthApi;
 //                 null if the request is not signed by anyone
 //                 or will be a string cartodb username otherwise.
 //
-AuthApi.prototype.authorizedBySigner = function(req, callback) {
-    if ( ! req.params.token || ! req.params.signer ) {
+AuthApi.prototype.authorizedBySigner = function(locals, callback) {
+    if ( ! locals.token || ! locals.signer ) {
         return callback(null, false); // no signer requested
     }
 
     var self = this;
 
-    var layergroup_id = req.params.token;
-    var auth_token = req.params.auth_token;
+    var layergroup_id = locals.token;
+    var auth_token = locals.auth_token;
 
     this.mapStore.load(layergroup_id, function(err, mapConfig) {
         if (err) {
@@ -86,7 +86,7 @@ AuthApi.prototype.authorizedByAPIKey = function(user, req, callback) {
  * @param req - standard req object. Importantly contains table and host information
  * @param callback function(err, allowed) is access allowed not?
  */
-AuthApi.prototype.authorize = function(req, callback) {
+AuthApi.prototype.authorize = function(req, res, callback) {
     var self = this;
     var user = req.context.user;
 
@@ -101,11 +101,11 @@ AuthApi.prototype.authorize = function(req, callback) {
             // if not authorized by api_key, continue
             if (!authorized)  {
                 // not authorized by api_key, check if authorized by signer
-                return self.authorizedBySigner(req, this);
+                return self.authorizedBySigner(res.locals, this);
             }
 
             // authorized by api key, login as the given username and stop
-            self.pgConnection.setDBAuth(user, req.params, function(err) {
+            self.pgConnection.setDBAuth(user, res.locals.db, function(err) {
                 callback(err, true); // authorized (or error)
             });
         },
@@ -120,7 +120,7 @@ AuthApi.prototype.authorize = function(req, callback) {
                 // if no signer name was given, let dbparams and
                 // PostgreSQL do the rest.
                 //
-                if ( ! req.params.signer ) {
+                if ( ! res.locals.signer ) {
                     return callback(null, true); // authorized so far
                 }
 
@@ -128,7 +128,7 @@ AuthApi.prototype.authorize = function(req, callback) {
                 return callback(null, false);
             }
 
-            self.pgConnection.setDBAuth(user, req.params, function(err) {
+            self.pgConnection.setDBAuth(user, res.locals.db, function(err) {
                 req.profiler.done('setDBAuth');
                 callback(err, true); // authorized (or error)
             });
--- a/lib/cartodb/api/auth_api.js
+++ b/lib/cartodb/api/auth_api.js
@@ -19,22 +19,22 @@ function AuthApi(pgConnection, metadataBackend, mapStore, templateMaps) {
 
 module.exports = AuthApi;
 
-// Check if a request is authorized by a signer
+// Check if the user is authorized by a signer
 //
-// @param req express request object
+// @param res express response object
 // @param callback function(err, signed_by) signed_by will be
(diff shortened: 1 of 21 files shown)


lib/cartodb/middleware/context/locals.js
```diff
@@ -0,0 +1,11 @@
+const _ = require('underscore');
+
+module.exports = function layergroupTokenMiddleware(req, res, next) {
+    // FIXME: Temporary hack to share data between middlewares. Express overrides req.params to
```
dgaubert: This comment is no longer valid.

test/unit/cartodb/prepare-context.test.js
```diff
@@ -43,74 +44,96 @@ describe('prepare-context', function() {
         assert.ok(_.isFunction(cleanUpQueryParams));
     });
 
-    function prepareRequest(req) {
+    function prepareRequest(req, res) {
```
dgaubert: Split this function in two. A function that returns two things is a bad smell.

lib/cartodb/middleware/context/locals.js
```diff
@@ -0,0 +1,11 @@
+const _ = require('underscore');
+
+module.exports = function layergroupTokenMiddleware(req, res, next) {
```
dgaubert: Bad function name

lib/cartodb/middleware/context/locals.js
```diff
@@ -0,0 +1,11 @@
...
+    // FIXME: Temporary hack to share data between middlewares. Express overrides req.params to
+    // parse url params to an object and it's performed after matching path and controller.
+    res.locals = {};
+    _.extend(res.locals, req.params);
+
+    next();
+};
```
dgaubert: Do we really need this middleware?

lib/cartodb/controllers/layergroup.js
```diff
@@ -198,9 +204,9 @@ LayergroupController.prototype.dataviewSearch = function(req, res, next) {
     step(
         function searchDataview() {
             var mapConfigProvider = new MapStoreMapConfigProvider(
-                self.mapStore, req.context.user, self.userLimitsApi, req.params
+                self.mapStore, req.context.user, self.userLimitsApi, res.locals
             );
-            self.dataviewBackend.search(mapConfigProvider, req.context.user, req.params, this);
+            self.dataviewBackend.search(mapConfigProvider, req.context.user, req.params.dataviewName, res.locals, this);
```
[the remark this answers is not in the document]
oleurud: Ok. I did not change it because we are not overwriting an Express property. But has sense to move it to res.locals

lib/cartodb/controllers/layergroup.js
```diff
@@ -315,37 +321,53 @@ LayergroupController.prototype.finalizeGetTileOrGrid = function(err, req, res, t
...
-        north: +req.params.north,
-        east: +req.params.east,
-        south: +req.params.south
-    }, next);
+    this.staticMap(
+        req, 
+        res, 
+        +req.params.width, 
+        +req.params.height, 
+        null,
```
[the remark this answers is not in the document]
oleurud: I will remove it, it is an error

lib/cartodb/controllers/named_maps.js
```diff
@@ -135,20 +135,20 @@ NamedMapsController.prototype.staticMap = function(req, res, next) {
...
             );
         },
         function prepareLayerVisibility(err, _namedMapProvider) {
             assert.ifError(err);
 
             namedMapProvider = _namedMapProvider;
-
-            self.prepareLayerFilterFromPreviewLayers(cdbUser, req, namedMapProvider, this);
+ 
+            self.prepareLayerFilterFromPreviewLayers(cdbUser, req, res, namedMapProvider, this);
```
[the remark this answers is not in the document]
oleurud: ok

lib/cartodb/middleware/context/authorize.js
```diff
@@ -1,15 +1,8 @@
...
     return function (req, res, next) {
-        // FIXME: Temporary hack to share data between middlewares. Express overrides req.params to
-        // parse url params to an object and it's performed after matching path and controller.
-        res.locals = {};
-        _.extend(res.locals, req.params);
-
         req.profiler.done('req2params.setup');
 
-        authApi.authorize(req, (err, authorized) => {
+        authApi.authorize(req, res, (err, authorized) => {
```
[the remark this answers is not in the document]
oleurud: ok

lib/cartodb/middleware/context/locals.js
oleurud: What is your proposal? Do you like something like tokenParserMiddleware?

lib/cartodb/middleware/context/locals.js
oleurud: ok

lib/cartodb/middleware/context/locals.js
oleurud: For me, his functionality is independent of the rest of middlewares and has no sense to mix with other. 

Follow your idea, maybe we can rename the "cleanUpQueryParamsMiddleware" to a more generic way and create the res.locals inside.

lib/cartodb/middleware/context/locals.js
enekid: is it even necessary to name the function? Would an arrow function be enough?
module.exports = (req, res, next) => {...}

test/unit/cartodb/prepare-context.test.js
oleurud: For me, it is not rare to find a function/module returning multiple values since destructuring assignment feature came with ES6. 

For me, it is ugly because I am returning the same objects (that are passed by reference). I created in this way to maintain the original idea, but maybe now the function is more visible.

Do you like something like:
- remove the return
- call prepareRequest without overwrite req and res ----->         prepareRequest(req, res);

lib/cartodb/middleware/context/locals.js
oleurud: Forgot my comment @dgaubert I did not see the file! Thanks

lib/cartodb/api/auth_api.js
```diff
@@ -26,15 +26,15 @@ module.exports = AuthApi;
 //                 null if the request is not signed by anyone
```
rochoa: Update comments accordingly.

lib/cartodb/api/auth_api.js
```diff
@@ -26,15 +26,15 @@ module.exports = AuthApi;
...
 //                 or will be a string cartodb username otherwise.
 //
-AuthApi.prototype.authorizedBySigner = function(req, callback) {
```
rochoa: I have the feeling we are trying to replace `req.params` with `res.locals`, but at the same time do some small refactors. This is just an example.

IMO, `AuthApi.authorizedBySigner` should receive the `token` and the `signer` because those are its real requirements.

However, as part of this PR (remember, change `req.params` with `res.locals`), we are changing (refactoring) the signature of this method: going from `req.params` to `locals` (we could call it `options`).

I would try to change just `req.params` with `res.locals` in this PR. Later, I would take the time to refactor small pieces like this one, removing completely the `req.params.*` and make clear what the signature/requirements for the function are.

lib/cartodb/api/auth_api.js
```diff
@@ -86,7 +86,7 @@ AuthApi.prototype.authorizedByAPIKey = function(user, req, callback) {
  * @param req - standard req object. Importantly contains table and host information
  * @param callback function(err, allowed) is access allowed not?
  */
-AuthApi.prototype.authorize = function(req, callback) {
+AuthApi.prototype.authorize = function(req, params, callback) {
     var self = this;
     var user = req.context.user;
```
rochoa: If, as proposed by @dgaubert, we put req.context.user into res.locals, we could just pass `res.locals` here (well, except for the `req.profiler` details), but again I have the feeling that we are trying to achieve two things here: one, replace `req.params` with `res.locals`; two, refactor.

For refactoring I would make really clear what the requirements are. I don't like the way we are introducing more dependencies into this `authorize` method. Actually, I find it more clear as `authorize(request)` than as `authorize(request, params)`. Thoughts?

lib/cartodb/backends/analysis-status.js
```diff
@@ -6,9 +6,7 @@ function AnalysisStatusBackend() {
 module.exports = AnalysisStatusBackend;
 
 
-AnalysisStatusBackend.prototype.getNodeStatus = function (params, callback) {
-    var nodeId = params.nodeId;
-
+AnalysisStatusBackend.prototype.getNodeStatus = function (nodeId, params, callback) {
```
rochoa: The `dbParamsFromReqParams` pattern happens all over the place: analyses[-status], dataviews, and turbo-carto.

If we are gonna start to be more granular with `res.locals`, I would consider start storing locals with some namespacing. For instance, I would have `res.locals.dbparams` where we store the result of `dbParamsFromReqParams(res.locals)`. So, later, we can pass res.locals.dbparams to methods like this, where we are only interested in the db connection params. Or even better, pass directly a PSQL collaborator.

Anyway, I think we should address this problem later, otherwise I'm falling into the same trap of confusing the replacement of `req.params` with the refactoring that enables.

lib/cartodb/backends/dataview.js
```diff
@@ -22,9 +22,7 @@ function DataviewBackend(analysisBackend) {
 
 module.exports = DataviewBackend;
 
-DataviewBackend.prototype.getDataview = function (mapConfigProvider, user, params, callback) {
-
-    var dataviewName = params.dataviewName;
+DataviewBackend.prototype.getDataview = function (mapConfigProvider, user, dataviewName, params, callback) {
```
rochoa: Same as in `analysis-status.js` comment.

lib/cartodb/controllers/analyses.js
```diff
@@ -44,7 +44,7 @@ AnalysesController.prototype.catalog = function (req, res, next) {
 
     step(
         function catalogQuery() {
-            var pg = new PSQL(dbParamsFromReqParams(req.params));
```
rochoa: Similar to `analysis-status.js` comment.

lib/cartodb/controllers/layergroup.js
```diff
@@ -315,37 +321,52 @@ LayergroupController.prototype.finalizeGetTileOrGrid = function(err, req, res, t
 };
 
 LayergroupController.prototype.bbox = function(req, res, next) {
-    this.staticMap(req, res, +req.params.width, +req.params.height, {
-        west: +req.params.west,
-        north: +req.params.north,
-        east: +req.params.east,
-        south: +req.params.south
-    }, next);
+    this.staticMap(
```
rochoa: Did we change anything here?

test/unit/cartodb/prepare-context.test.js
dgaubert: What about to have two functions like `prepareRequest` and `prepareResponse`?

lib/cartodb/middleware/context/locals.js
dgaubert: Following @rochoa's comments, let's first get done this PR and we'll think in further refactors afterward.

lib/cartodb/backends/analysis-status.js
oleurud: This was my first approach but is a big change because we need to change dependencies. 

As you are saying, we are doing some refactors (in addition to res.locals) and the idea is to maintain the scope small for this PR, but sure is a nice future improvement

lib/cartodb/api/auth_api.js
oleurud: I am not sure about what changes you are requesting here.

Of course, "locals" is a bad name, I am using "params" all the time, I don't know what I was thinking when I change this.

And yes, AuthApi.authorizedBySigner should receive the _token_, _signer_ and _auth_token_ and not an Express object. The real problem is to have the Express Request or Response objects in a service which doesn't need it and we can change _request_ by _response_, _locals_ by _params_ or _options_ and the real problem continues there

So, is the idea to receive the Response object and in other moment do the refactor?

lib/cartodb/controllers/layergroup.js
oleurud: Not really, only more readable. I will be back with this

lib/cartodb/api/auth_api.js
oleurud: Ok, I totally forgot it

lib/cartodb/controllers/layergroup.js
oleurud: Do you want to undo it?

lib/cartodb/middleware/context/locals.js
oleurud: @enekid Is a nice change, these names are not accessibles. But not for this PR, I will try in middlewarify PR (when I merge this)

oleurud:
How do you see it? @dgaubert @rochoa

lib/cartodb/controllers/layergroup.js
dgaubert: 🙏

lib/cartodb/api/auth_api.js
dgaubert: Let's follow @rochoa's suggestion, for this method I'd change `req` by `res` and then change `req.params` by `res.locals` in `authorizedBySigner`method.

lib/cartodb/backends/analysis-status.js
dgaubert: Agreed, let's face this later.

lib/cartodb/api/auth_api.js
oleurud: Ok

test/unit/cartodb/prepare-context.test.js
```diff
@@ -47,70 +48,91 @@ describe('prepare-context', function() {
...
-          if ( err ) { done(err); return; }
-          assert.ok(_.isObject(req.query), 'request has query');
-          assert.ok(!req.query.hasOwnProperty('dbuser'), 'dbuser was removed from query');
-          assert.ok(req.hasOwnProperty('params'), 'request has params');
-          assert.ok(!req.params.hasOwnProperty('interactivity'), 'request params do not have interactivity');
-          assert.equal(req.params.dbname, test_database);
-          assert.ok(req.params.dbuser === test_pubuser, 'could inject dbuser ('+req.params.dbuser+')');
-          done();
-      });
+        var req = {headers: { host:'localhost' }, query: {} };
```
dgaubert: In order to avoid noise, can we skip style format changes?

oleurud: ok, done

lib/cartodb/controllers/layergroup.js
oleurud: Done

lib/cartodb/api/auth_api.js
```diff
@@ -84,11 +84,12 @@ AuthApi.prototype.authorizedByAPIKey = function(user, req, callback) {
  * Check access authorization
  *
  * @param req - standard req object. Importantly contains table and host information
+ * @param params res.locals parameters. Contains the auth parameters
  * @param callback function(err, allowed) is access allowed not?
  */
-AuthApi.prototype.authorize = function(req, callback) {
+AuthApi.prototype.authorize = function(req, res, callback) {
```
dgaubert: Why do we need `req` here? After moving `req.context.user` to `res.locals.user`we don't use req anymore. Do we?

oleurud: For req.profiler and authorizedByAPIKey method (now needs to access to req.query and req.body)

dgaubert: ok! I did not see it.

lib/cartodb/api/auth_api.js
```diff
@@ -84,11 +84,12 @@ AuthApi.prototype.authorizedByAPIKey = function(user, req, callback) {
  * Check access authorization
  *
  * @param req - standard req object. Importantly contains table and host information
+ * @param params res.locals parameters. Contains the auth parameters
```
rochoa: Typo.

lib/cartodb/backends/analysis-status.js
rochoa: So, are we keeping things like they were or are we willing to keep some of this small refactors?

test/acceptance/ported/support/ported_server_options.js
```diff
@@ -74,6 +74,9 @@ module.exports = _.extend({}, serverOptions, {
         }
         req.params.dbname = 'test_windshaft_cartodb_user_1_db';
 
+        // add all params to res.locals
+        res.locals = _.extend({}, req.params);
```
rochoa: Isn't a bit dangerous to put everything in here before cleaning up in `lib/cartodb/middleware/context/clean-up-query-params.js`?

lib/cartodb/backends/analysis-status.js
oleurud: Ok, I will undo this, it is not the PR to do it

lib/cartodb/server.js
```diff
@@ -369,7 +369,8 @@ function bootstrap(opts) {
     app.use(bodyParser.json());
 
     app.use(function bootstrap$prepareRequestResponse(req, res, next) {
-        req.context = req.context || {};
+        res.locals = {};
```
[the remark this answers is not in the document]
oleurud: Yeep, but not in the way as we expect. 

I don't know the reason, but res.locals does not have the Object.prototype. More info: https://github.com/expressjs/express/issues/2771

Your this reason, I used this easy workaround

test/acceptance/ported/support/ported_server_options.js
oleurud: I am a little lost with this function, but I can explain it 

In some cases (this is that I don't know well) this method replaces all the new middlewares (which replaces the old req2params from the base controller). You can see this replacement in line 214 of server.js
```
const prepareContext = typeof serverOptions.req2params === 'function' ?
        serverOptions.req2params :
        prepareContextMiddleware(authApi, pgConnection);
```

So, when the code uses this method, not uses the lib/cartodb/middleware/context/clean-up-query-params.js

lib/cartodb/server.js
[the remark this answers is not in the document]
oleurud: Yeep, but not as we expect.

The res.locals object does not have the Object.prototype. So, for example, we could not use a method like hasOwnProperty (and we use it). More info: https://github.com/expressjs/express/issues/2771

So, I decided to use this easy workaround

lib/cartodb/api/auth_api.js
oleurud: Sorry, I forget again. Done

test/acceptance/ported/support/ported_server_options.js
dgaubert: AFAIK, `serverOptions.req2params` is intended for testing purposes and replaces the old `req2params` method and from now on the new middlewares.

lib/cartodb/server.js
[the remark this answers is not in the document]
dgaubert: Do we really need to use `hasOwnProperty`??  If so, which I think we don't,  there is an alternative: `Object.prototype.hasOwnProperty.call(res.locals, 'wadus')`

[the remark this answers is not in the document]
oleurud: Is not necessary, I never use it. 

But, to change this (and another possible things) is a bigger refactor than:
```
res.locals = {};
```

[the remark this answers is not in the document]
dgaubert: If I'm not wrong we only use `res.locals.hasOwnProperty('token')` [here](https://github.com/CartoDB/Windshaft-cartodb/blob/res-locals/lib/cartodb/middleware/context/layergroup-token.js#L4)

So, it is not a hard thing. is it?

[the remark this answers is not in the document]
oleurud: I am committing removing the overwrite res.locals creation and changing the check that you say @dgaubert 

It was the only point where the res.locals was affected by Object.prototype and the tests are working

lib/cartodb/backends/analysis-status.js
oleurud: I want to explain a little more. How I said in the previous comment: "the idea was to use req.params to get the nodeId (is an URL parameter)". 

So, not to mix the function of req.params and res.locals and ideally, IMHO we must not have the req.params in res.locals. I have tried to remove this, but is not possible in this PR, because in some libraries we are using the URL params from res.locals

oleurud:
Please @rochoa, can you review this PR again? :)

PR #766, opened 2017-10-03
Update yarn version to v0.27.5

rafatower:
Update the engine requirements. The file `yarn.lock` has minor changes (omitting the github prefix, as a result of my first `yarn install`).

This fixes the (yarn itself) installation problem and keeps it consistent with travis.

cc'ing @dgaubert for awareness.


package.json
```diff
@@ -62,6 +62,6 @@
   },
   "engines": {
     "node": ">=6.9",
-    "yarn": "^0.21.3"
+    "yarn": "^0.27.5"
```
rochoa: I don't know if this format plays well with yarn. Example:

```shell
$ yarn --version
0.28.4
$ yarn
yarn install v0.28.4
$ make pre-install
[1/5] 🔍  Validating package.json...
error windshaft-cartodb@3.13.1: The engine "yarn" is incompatible with this module. Expected version "^0.27.5".
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

However, with:

```diff
$ git diff
diff --git a/package.json b/package.json
index 6532a04e..fd857e96 100644
--- a/package.json
+++ b/package.json
@@ -62,6 +62,6 @@
   },
   "engines": {
     "node": ">=6.9",
-    "yarn": "^0.27.5"
+    "yarn": ">=0.27.5 <1.0.0"
   }
 }
```

It works:

```shell
$ yarn
yarn install v0.28.4
$ make pre-install
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
success Already up-to-date.
```

Should we go with something like:

rafatower: [In theory it should work](https://yarnpkg.com/lang/en/docs/dependency-versions/#toc-caret-ranges) (at least in newer versions) but as it break... changing it...

PR #767, opened 2017-10-05
Implement stats middleware removing some duplicated code

dgaubert:


lib/cartodb/middleware/stats.js
```diff
@@ -0,0 +1,33 @@
+const Profiler = require('../stats/profiler_proxy');
+const debug = require('debug')('windshaft:cartodb:stats');
+const methods = ['send', 'json', 'jsonp'];
+
+module.exports = function statsMiddleware(options) {
```
dgaubert: With Koa it'd be resolved as follows:
```js
module.exports = function statsMiddleware({ enabled = true, statsClient }) {
    return async function stats(ctx, next) {
        ctx.profiler = new Profiler({
            statsd_client: statsClient,
            profile: enabled
        });

        await next();

        ctx.set('X-Tiler-Profiler', ctx.profiler.toJSONString());

        try {
            ctx.profiler.sendStats();
        } catch (err) {
            debug("error sending profiling stats: " + err);
        }
    };
};
```
:trollface:

dgaubert:
Hey!

I've improved a bit this PR. The goal is to use `on-headers` and `finish` event to avoid to decorate all functions we use to respond the request. Well, to be honest `on-headers` decorates `res.writeHead` method, but is just one method instead of three.

Please, a CR would be appreciated.

oleurud:
For me, the idea is nice and makes sense: remove duplicated code and do the stats things in a file with a semantic valor.

PR #773, opened 2017-10-16
Fix tests in master branch while testing in devel environment (NOT CI)

dgaubert:
Tests in CI are using a docker image in order to install new stuff about MVT (Postgis 2.4). But running them in development simply fail. 

Now, test related to the new MVT renderer run when POSTGIS_VERSION environment variable is present and equal to '2.4' otherwise these tests are skipped.


enekid:
:eu:

PR #775, opened 2017-10-16
added map tile rendering and MVT info to Maps API doc- WIP

csobier:
Fixes https://github.com/CartoDB/docs/issues/1236

@inigomedina, I know you said to keep this high-level for now so I'm not sure if it needs full examples? Also, I wasn't sure if I am confusing the "Vector rendering" feature with the "MVT" feature, but I think they are one in the same. This is a just rough draft to start, please review.

PR #778, opened 2017-10-18
mvt content added to anonymous map chapter, under retrieve resources …

csobier:
…heading- WIP

@inigomedina , necessary edits required for the initial commit https://github.com/CartoDB/Windshaft-cartodb/pull/775. 

Fixes https://github.com/CartoDB/docs/issues/1236. 

**Note**: There is now no mention of the related PostGIS function for MVT files. We also still need information about MVT limitations and feature specific properties that Santana mentioned (such as buffer, resolution, etc.). Let me know if this requires some additional MapConfig content and/or a better strategy for how we share this info.

### DIFF
--- a/docs/anonymous_maps.md
+++ b/docs/anonymous_maps.md
@@ -70,21 +70,203 @@ curl 'https://{username}.carto.com/api/v1/map' -H 'Content-Type: application/jso
 }
 ```
 
-### Retrieve resources from the layergroup
+## Map Tile Rendering
+
+Map tiles create the graphical representation of your map in a web browser. The performance rendering of map tiles is dependent on the type of geospatial data model (raster or vector) that you are using.
+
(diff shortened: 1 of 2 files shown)


docs/anonymous_maps.md
```diff
@@ -70,21 +70,203 @@ curl 'https://{username}.carto.com/api/v1/map' -H 'Content-Type: application/jso
...
+```
+
+5. Request Tiles (from CARTO) and Set to Map Object (Mapbox):
+
+**Note:** By default, [CARTO core functions](https://carto.com/docs/carto-engine/carto-js/core-api/) retrieve URLs for fully rendered tiles. You must replace the default format (.png) with the MVT format (.mvt).
+
+
+```bash
+cartodb.Tiles.getTiles(layerOptions, function(result, err) {
+var tiles = result.tiles.map(function(tileUrl) {
```
javisantana: arg, don't you think a small update on core library to add this would make this much much better?

PR #779, opened 2017-10-20
Don't overwrite data in copy from req.params to res.locals

oleurud:
Now, we are ensuring that req.params.user not overwrite res.locals.user in usermiddleware. But probably this middleware is not the correct place to do it.

Also, we are not checking if others properties are being overwritten


oleurud:
do you like? @dgaubert ;)

PR #784, opened 2017-11-02
[Middlewarify] Middlewarify map controller

dgaubert:
Still, in development. **Don't ready for CR**. Just for awareness. 

**Note**: I've tried as much as possible to commit step by step in order to show you the process that I've followed.

This PR aims to break responsibilities into middlewares. Each middleware tries to follow SRP principle. By composing them in a set of middlewares we can perform, for instance, map creation (anonymous) or map instantiation (named map).

The good parts:
  - Smaller functions, improving readability and maintainability (easy to add, modify or remove middlewares). We will be able to compose and reuse them in multiple ways.
  - Improved error handling: taking advantage of middleware system to short-circuit the normal workflow.
  - Graceful asynchronous code handling: by using callback provided by express' middleware system. Step module is no longer needed.

The bad parts:
  - Rely on `res.locals` to pass info and results of previous middlewares to next middleware. I do believe there is a room of improvements here. Stay tuned ;)

This PR materializes the middleware architecture that I've tried to explain some time ago.


dgaubert:
Hey folks!

It's alive!!! and ready for CR...

I've stopped this PR at this point, we can go deeper in some areas but I thought that it could be too hard to follow the following changes. I've broken functionality as much as possible, by refactoring other controllers we will find common middlewares among them. At the moment, all middlewares belong to map controller but my idea is that they end up being refactored as isolated middlewares. 

Furthermore, map controller is responsible for building middlewares, so map controller needs their dependencies to build them. In the future controller should only know about middlewares. After refactoring all controllers, I want to see how the rest of them (layergroup, named-maps, etc..) look like before changing current app building system.

Please don't hesitate to comment or share your comments.

enekid:
`res.locals` is crying out loud for a refactor. Currently it looks like a catchall global variable. It's better than before, though.

We need to order `res.locals` into buckets: 

- res.locals.db
- res.locals.auth
- res.locals.user
...

In order to abstract this res.locals thing, we could initialize all the middlewares with the appropriate context buckets. 

```javascript
   MapController.prototype.composeCreateMapMiddleware = (res, useTemplate = false) => {
   ...
   const context = res.locals;

    return [
        cors(),
        userMiddleware(context),
        this.prepareContext(context),
        doSomethingWithUser({user: context.user}),
        doSomethingWithUserAndAuth({user: context.user, auth: context.auth})
    ];
};
```

I think that this `res.locals` refactor should NOT go in this PR.

dgaubert:
> `res.locals` is crying out loud for a refactor.
 
+1000

> In order to abstract this `res.locals` thing, we could initialize all the middlewares with the appropriate context buckets

If we go for it then we have to compose middlewares for each incoming request. That's is going to penalize response times for sure. So we need to create middleware chain while building the application.

For the time being, I'll go for your proposal of namespacing `res.locals`. 

Thanks for the feedback

lib/cartodb/controllers/layergroup.js
```diff
@@ -401,43 +399,48 @@ LayergroupController.prototype.sendResponse = function(req, res, body, status, h
...
             }
+        } else {
+            res.send(body);
         }
-    );
+    });
+};
 
+LayergroupController.prototype.setCacheChannel = function(req, res, callback) {
+    const { dbname, user, token } = res.locals;
```
oleurud: We must try to stop using the URL params from res.locals and start using them from req.params. Is the first step to decouple both "params containers" and in the future we can remove req.params from res.locals

oleurud:
:+1: totally agree with your comments

We must try to stop using the URL params from res.locals and start using them from req.params. Is the first step to decouple both "params containers" and in the future we can remove req.params from res.locals

enekid:
The final objective should be to establish `res.locals` as the single source of truth for the input parameters. `res.locals` should be initialized by the middlewares with the data held in (among others):
- res.params
- res.body
- res.query
- res.get(headerName)
- res.hostname

The controllers should work only with `res.locals`.

oleurud:
I don't think so or maybe I'm not understanding you.

Why do we need to duplicate de params? What is the advantage of reading a URL param from res.locals instead of req.params or a header from res.locals instead of req.get()? Why doesn't a controller work with Request and Response objects (including the res.locals)?

In addition to "maintaining the same things twice" and the problems related to it, using a single source of truth adds a lot of problems: services/methods receiving unneeded info, lazy dependencies ... Really, until some weeks ago, we had a single source of truth: req.params

enekid:
The controllers shouldn't know that they run inside an Express application (req/res). They should receive only a big context parameter (set to res.locals in this case) or ideally an object with only the data they need: {user, db, ...}

We refactored/are refactoring it to res.locals instead of req.params because it's the recommended way of doing it now [(stackoverflow)](https://stackoverflow.com/questions/33451053/req-locals-vs-res-locals-vs-res-data-vs-req-data-vs-app-locals-in-express-mi). `req.params` should only be used for the request's urls params (/:i-am-a-parameter)

One purpose of using res.locals is to abstract the controllers of the info's source. For example some parameters can be sent using query params, body fields or http headers. Handling this in the controllers only adds noise. In addition, this approach let's us decouple input handling from processing.

Nice thread 😄

enekid:
what I said, but in [animated gif format](https://github.com/i0natan/nodebestpractices/blob/master/sections/projectstructre/createlayers.md)

Issue #782, opened 2017-10-31
Return custom mvt when limits are reached


rubenmoya:
We need to return a custom mvt with a `429` status code when limits are reached. 

Design issue: #TBD

dv343:
This requires to
 - Pre-generate an error MVT tile with the orange stripped square
 - Return that static tile on error

oleurud:
Development finished and working on staging

dv343:
Isn't this done already?

oleurud:
It is not in production. Maybe you will need to wait until Monday

rubenmoya:
We will start merging things tomorrow! I'll keep you posted!

rubenmoya:
Closed via https://github.com/CartoDB/Windshaft-cartodb/pull/787

PR #787, opened 2017-11-07
[DNM] Timeout vector image

oleurud:
Related issue: #782
When a timeout error happens in vector, we must return a 429 error and an mvt error tile.


test/acceptance/user-database-timeout-limit.js
```diff
@@ -411,26 +411,31 @@ describe('user database timeout limit', function () {
...
+                    this.testClient.getTile(0, 0, 0, params, (err, res, tile) => {
                         assert.ifError(err);
-
-                        assert.deepEqual(timeoutError, DATASOURCE_TIMEOUT_ERROR);
-
+    
+                        var tileJSON = tile.toJSON();
+                        assert.equal(Array.isArray(tileJSON), true);
+                        assert.equal(tileJSON.length, 2);
+                        assert.equal(tileJSON[0].name, 'errorTileSquareLayer');
```
dgaubert: We use to use a helper to compare tiles (`raster` so far) but as `vector` is going to be mainstream It'd be really nice to have a function helper to compare vector tiles directly, as we do already here: https://github.com/CartoDB/Windshaft-cartodb/blob/master/test/support/assert.js#L14-L38

We can use [VectorTile.render()](http://mapnik.org/documentation/node-mapnik/3.6/#VectorTile.render) and then compare images. What do you think??

lib/cartodb/controllers/layergroup.js
```diff
@@ -67,7 +70,8 @@ LayergroupController.prototype.register = function(app) {
         userMiddleware,
         validateLayerRouteMiddleware,
         this.prepareContext,
-        this.layer.bind(this)
+        this.layer.bind(this),
+        vectorError()
```
dgaubert: let's see how it grows, but having such a expecific middleware to handle error we will end up having something like this:

```
app.get(
   [...],
   this.layer.bind(this),
   vectorError(),
   rasterError(),
   limitsError(),
   wadusError()
);
```
For each route, wich is kind of boilerplate. I think we will have to use a middleware builder to compose all error middlewares:

```
app.get(
   [...],
   this.layer.bind(this),
   layerError()
);

// layerError.js
module.exports = function layerError (options) {
  return function layerErrorMiddlewre (req, res, next) {
    return [
      vectorError(),
      rasterError(),
      limitsError(),
      wadusError()
    ]
  }
}
```

oleurud: I thought about it, I decided to do the easiest code for this case but start the [error middleware](https://github.com/CartoDB/Windshaft-cartodb/pull/789) discussion.

dgaubert: Hey! 

It was just a heads-up, not to be implemented in this PR. In fact, I like the simplicity of this PR. 

Sorry for not being clear enough.

oleurud: Ok, understood

test/acceptance/user-database-timeout-limit.js
oleurud: Sounds well, but I am not sure about it because the vector tile is not a pixels chain to be compared. There are no colors and channels, only geometries. I think is more like object comparison or a JSON comparison. 

In any case, I like the idea of having a method to compare mvt files. Probably @dv343 can give us a good point of view.

dv343: IMHO I don't like the idea of rendering the MVTs for multiple reasons:
* Test times will be unnecessarily high
* You'll need mapnik as a dependency (I know we already have it, but we shouldn't need it for vector)
* Windshaft has little room for small errors with MVT. With the last St_AsMVT it just forwards the DB output. There is no style inside the MVTs, all style related errors can only happen at frontend.
* Comparing bit by bit (or by hash) is fine. Only the DB should be capable of changing the result, and if does it should be because A) you actually changed the query or B) the DB has applied a different simplification / winding order / validity enforcement. A) would mean an error in Windshaft and B) could potentially break things on frontend, so, better to know, and if it fails we better should know why.

oleurud: Thanks @dv343 Anyway, I wouldn't like to do a big change in this PR and I prefer to create another to do it. What do you think @dgaubert ?

dv343: BTW, another option is to decode the MVT and check for a few features, this is already done in some tests.

oleurud: I will make another PR to make the helper function to compare vector tiles because we need to deploy this feature.

Kill me if I won't create it today

PR #798, opened 2017-11-20
[WIP] New auth system

dgaubert:


lib/cartodb/middleware/auth/authorize.js
```diff
@@ -0,0 +1,77 @@
...
+      return next(createError(401, 'api-key-required', {type: 'auth', subtype: 'api-key-required'}));
+    }
+
+    if ( ! contextAndApiKeyUsernamesMatch(context)) {
+      //TODO too much information, sent instead api-key-not-found. IS THIS REALLY NECESARRY?
+      return next(createError(401, 'api-key-username-mismatch', {type: 'auth', subtype: 'api-key-mismatch'}));
+    }
+
+    fetchApiKeyFromMetadataBackend({metadataBackend, username, apiKeyToken}, (err, apiKey) => {
+      if ( err || ! apiKey) {
```
dgaubert: I'd differentiate when there is an error (due to any problem with Redis) from the fact that api-key was not found.

lib/cartodb/server.js
```diff
@@ -67,7 +67,7 @@ module.exports = function(serverOptions) {
         global.statsClient.gauge(keyPrefix + 'waiting', status.waiting);
     });
 
-    var metadataBackend = cartodbRedis({pool: redisPool});
+    var metadataBackend = cartodbRedis({pool: redisPool, apiName: 'maps'});
```
dgaubert: Metadata should not keep this info at creation level, it's better to pass this info to `getApiKey()`.

dgaubert:
Hey @enekid,

I'd like to share some thoughts after reviewing the PR.

For sure we should put some effort to get `res.locals` healthy by applying namespacing or whatever. However, I'm afraid that we are doing it in the wrong order. 

Let me share with you some context about why `res.locals` is a mess nowadays. Before using `res.locals`, we had the pretty same for `req.params` we used it as object container to have everything related to path params (e.g. tile coordinates, etc..), query params (e.g. `api_key` etc..) and user's database connection params (e.g. `dbuser`, `dbhost`, etc..) in the same place. Then we pass `req.params` to many methods as a parameter and they use it, and worse, modify it. A clear example of that is the _current_ `auth-api` that receives `req.params` to read the user and api-key and populates it with database connection params.

As you already know, by using middlewares more and more we moved `req.params` to `res.locals` to give us the chance of sharing more info among middlewares so we put more stuff into it. 

The following are some examples of `res.locals` looks like:
```js
// Map creation (anonymous)
{ 
  user: 'localhost',
  api_key: '1234',
  dbuser: 'test_windshaft_cartodb_user_1',
  dbpassword: 'test_windshaft_cartodb_user_1_pass',
  dbhost: 'localhost',
  dbport: 5432,
  dbname: 'test_windshaft_cartodb_user_1_db' 
}
// Get tile
{ 
  user: 'localhost',
  token: 'fb0051fc202bae9b064a5f9bd6b3672a',
  z: '14',
  x: '8023',
  y: '6177',
  format: 'png',
  api_key: '1234',
  cache_buster: '0',
  dbuser: 'test_windshaft_cartodb_user_1',
  dbpassword: 'test_windshaft_cartodb_user_1_pass',
  dbhost: 'localhost',
  dbport: 5432,
  dbname: 'test_windshaft_cartodb_user_1_db' 
}
// Map instantiation with GET
{ 
  user: 'localhost',
  config: '{"version":"1.5.0","layers":[{"options":{"source":{"id":"2570e105-7b37-40d2-bdf4-1af889598745"},"sql":"select * from test_table limit 2","cartocss":"#layer { marker-fill:red; }","cartocss_version":"2.3.0","attributes":{"id":"cartodb_id","columns":["name","address"]}}}],"analyses":[{"id":"2570e105-7b37-40d2-bdf4-1af889598745","type":"source","params":{"query":"select * from test_table limit 2"}}]}',
  api_key: '1234',
  dbuser: 'test_windshaft_cartodb_user_1',
  dbpassword: 'test_windshaft_cartodb_user_1_pass',
  dbhost: 'localhost',
  dbport: 5432,
  dbname: 'test_windshaft_cartodb_user_1_db' 
}
```
By applying namespaces to `res.locals` we can resolve the problem partially. But first of all, we need to refactor every function that receives `res.locals` and change the way that reads from it. Now, test are broken because we put some info in `req.locals.auth` or `req.locals.db` but functions that relies on `res.locals` doesn't know how to use it.

As I said before, just applying namespaces to `res.locals` we don't fix the real problem. Actually, we are missing what `req.params` and `req.query` are intended for. If any function (or class method) needs info from `req.params` and `req.query` we should provide only the info that they need instead of the whole `req.locals`, for instance:

```js
// Get tile
req.params = { 
  token: 'fb0051fc202bae9b064a5f9bd6b3672a',
  z: '14',
  x: '8023',
  y: '6177',
  format: 'png'
};
req.query = {
  api_key: '1234',
  cache_buster: '0'
};
res.locals = {
  user: 'localhost',
  dbuser: 'test_windshaft_cartodb_user_1',
  dbpassword: 'test_windshaft_cartodb_user_1_pass',
  dbhost: 'localhost',
  dbport: 5432,
  dbname: 'test_windshaft_cartodb_user_1_db' 
};

// before
new MapStoreMapConfigProvider(mapStore, res.locals.user, userLimitsApi, res.locals);
// after
const options = {
  token: req.params.token,
  cache_buster: req.query. cache_buster,
  apy_key: req.query.api_key,
  dbname: req.locals.dbname
  dbuser: req.locals.dbuser,
  [...]
}
const mapConfigProvider = new MapStoreMapConfigProvider(mapStore, res.locals.user, userLimitsApi, options)
```
It's more verbose but clear, less "clever" but easy to follow. Don't you agree?

Once we have `req.locals` less populated we can apply some namespacing, for instance:
```
res.locals = {
  user: {
    name: 'localhost',
    connectionParams: {
      user: 'test_windshaft_cartodb_user_1',
      pass: 'test_windshaft_cartodb_user_1_pass',
      host: 'localhost',
      port: 5432,
      dbname: 'test_windshaft_cartodb_user_1_db'
   },
   [...]
};
```
As you can see the problem we have with `res.locals` is bigger than we thought initially. From my point of view, it's out of the scope of the implementation of new auth system but we can move towards a healthy `res.locals`.

Hope this helps.

**Don't hesitate to ask questions about that or any doubt that you have**

dgaubert:
I've spent some time reviewing what are the pieces affected by the new brand auth-api, in order to give us more context before going forward:

To gather the right connection params for the user we use two different middlewares:
 - **Authorize middleware** (authorize.js): checks if the given `api_key` or `auth_token` are valid. If so, sets user's `dbuser` & `dbpassword` to `res.locals` from user's metadata.
 - **DB connection setup middleware** (db-conn-setup.js): populates `res.locals` with database connection params: `dbhost` & `dbname` from user's metadata and populates `dbuser` & `dbpassword` with `publicuser` and `default-password` only if _authorize middleware_ didn't do it (because neither `api_key` nor `auth_token` were provided we use the well known `publicuser`).

Both middlewares rely on `PgConnection` class. This class provides two methods to populate connection params:
 - `.setDBAuth(user, params, callback)`: populates argument `params` with user's `dbuser` and `dbpassword` if `params` doesn't have them yet.
 - `.setDBConn(user, params, callback)`: populate argument `params` with `dbhost`, `dbname`, `dbport` and if `params` doesn't have `dbuser` & `dbpassword` yet then sets them to `publicuser` & `default-password`.

So `PgConnection` is used to get connection parameters and set them to `res.locals` or whatever object we want to pass. For instance, we can get `publicuser`'s connection params:
```js
const user = 'rambo'
const params = {}
pgConnection.setDBConn(user, params, err => {
  if (err) return console.error(err);
  console.log(params);
})
```

Or we can get `user`'s connection params:
```js
const user = 'rambo'
const params = {}
pgConnection.setDBAuth(user, params, err => {
  if (err) return console.error(err);
  pgConnection.setDBConn(user, params, err => {
    if (err) return console.error(err);
    console.log(params);
  })
})
```

And last but not least, we can get a `user`'s database connection directly by using `.getConnection(username, callback)` method:

_it does something similar that I described before_
```js
const user = 'rambo';
pgConnection.getConnection(user, (err, connection) => {
  if (err) return console.error(err);
  connection.query('select * from private_table', (err, res) => {
    [...]
  })
});
```

Having `res.locals` fulfilled with connection params (publicuser or user) we can create a PSQL (our custom PgClient):
```js
const psql = new PSQL({
  user: res.locals.dbuser,
  pass: res.locals.dbpass,
  host: res.locals.dbhost,
  port: res.locals.dbport,
  dbname: res.locals.dbname
});
connection.query('select * from populated_places', (err, res) => {
  [...]
})
```

The whole point is: both `pgConnection` and `res.locals` are injected as dependencies to controllers (layergroup, map, etc..), models (mapconfig-providers, dataviews, etc..), backends (tile, stats, etc..). 

**They are heavily used**.❗️❗️❗️

To sum up: we have to be careful changing anything related to `res-local`

dgaubert:
## Where do we pass database credentials to `mapnik` to render a tile?
Through `MapConfigProvider`. In [`LayergroupController`](https://github.com/CartoDB/Windshaft-cartodb/blob/master/lib/cartodb/controllers/layergroup.js#L283-L286) we create a `MapStoreMapConfigProvider` which saves the database connection params. That `MapConfigProvider` is passed to the [`TileBackend`](https://github.com/CartoDB/Windshaft/blob/master/lib/windshaft/backends/tile.js#L24-L64) as an argument and in `Windshaft` we use it by calling [`.getMapConfig()`](https://github.com/CartoDB/Windshaft/blob/master/lib/windshaft/cache/renderer_cache.js#L80-L85) which returns the credentials and are given to the `MapnikFactory`. Finally, the [renderer](https://github.com/CartoDB/Windshaft/blob/master/lib/windshaft/renderers/mapnik/factory.js#L162) uses that credentials to give them to [`Grainstore`](https://github.com/CartoDB/grainstore) which is in charge of generating the `xml` that `Mapnik` needs to load a map and return the tile.

enekid:
Excellent comments!

Now mine :)
 
- `res.locals` should be used extensively in the middlewares. All the `req.params`, `req.query`, `req.body`, `req.headers`, etc should be merged into `res.locals` along other information gathered/generated by the middlewares (API Keys, Rate Limiting, etc). `res.locals` should be the central source of truth. That said, the controllers/libraries that make the heavy stuff should not receive the whole `res.locals` object. As you said, we should provide them only the information they need, creating a new options object.
- The _everything to a namespaced_`res.locals` approach forces a big refactor of the other components that expect a _non namespaced_ `res.locals`. We should postpone the complete namespaced `res.locals`. But I thing we should consider to start it know, at least with `res.locals.db `and `res.locals.auth`  
- We should get rid of **setDBAuth** and **setDBConn**. That's something that the db middleware should do. `PgConnection.getConnection` should receive the `res.locals.db` and a flag indicating if we want to connect with the master credentials. Probably better, split the functionality in 2 functions: `getConnection` and `getMasterConnection`.

dgaubert:
Excellent, you really got it! 💪. Point by point:

1) I cannot agree more. 
2) For sure, you will find some weird behaviors and broken tests in ways that you won't expect. If you feel confident with that, GO AHEAD!
3) _PgConnection.getConnection should receive the res.locals.db and a flag indicating if we want to connect with the master credentials. Probably better, split the functionality in 2 functions: getConnection and getMasterConnection._ That's a possibility that I like. Looking forward to seeing how it evolves.

dgaubert:
Superseded by #877

PR #800, opened 2017-11-24
Logging all errors

oleurud:
We need to log more info when the app returns an error (limits types, errors in JSONP, ...)

Following the current way, we think to use the log4js and adding a new header with the errors info.

The same task in https://github.com/CartoDB/CartoDB-SQL-API/pull/451


oleurud:
I have followed the same structure and we will need to refactor it in https://github.com/CartoDB/Windshaft-cartodb/pull/789

The final log:
```
[2017-11-27 16:06:58.172] [INFO] [default] - undefined HEAD localhost:6081/ 200 2 ms -> text/html; charset=utf-8 ({}) !!!{}

or

[2017-11-27 17:14:22.726] [INFO] [default] - undefined GET cdb.localhost.lan:8181/api/v1/map/6a38d0caaf7d53af1cfa09252d603e76:1509970291768/1/2/2/0.grid.json?api_key=7c0df56a8759a42fd1306019d4de7c5a9e784d7a&callback=grid 200 108 ms -> text/javascript; charset=utf-8 ({"authorizedByAPIKey":2,"authorize":3,"req2params":1,"getTileOrGrid":98,"getRenderer":50,"render-grid-json":48,"render":47,"encode":0,"total":6}) !!!{"statusCode":429,"message":"You are over platform's limits. Please contact us to know more details","name":"Error","type":"limit","subtype":"datasource","moreErrors":[]}
```

It is ready for CR

lib/cartodb/middleware/error-middleware.js
```diff
@@ -160,3 +161,31 @@ function errorMessageWithContext(err) {
 
     return error;
 }
+
+function logErrors(errors, statusCode, res) {
```
dgaubert: Can we use a name more descriptive for this function? For instance:  `setErrorHeader`

lib/cartodb/server.js
```diff
@@ -409,6 +409,12 @@ function setupLogger(app, opts) {
         };
         app.use(global.log4js.connectLogger(global.log4js.getLogger(), _.defaults(loggerOpts, {level: 'info'})));
     }
+
+    // default X-Tiler-Errors header
+    app.use((req, res, next) => {
```
dgaubert: Why do we need to initialize this header? It's wide better to not set it, isn't it? we have to make this part optional for the Regex that consumes the log because the 99% of the time we don't log errors.

test/unit/cartodb/base_controller.js
```diff
@@ -20,4 +20,106 @@ describe('error-middleware', function() {
             "Error status code for multiline/PSQL does not match"
         );
     });
+
```
dgaubert: 😱 This is a vestigial test suite from the old `base_controller`. Please, rename this file: base_controller.js => error-middleware.test.js or something like that. 🙏

lib/cartodb/server.js
oleurud: Since the header is added to environment config file, it will appear in the logs. We don't set a default value, the log will be `!!!undefined`. I dont like it, and this is the reason why I set a default value to have `!!!{}` instead of

oleurud: Finally, I have changed it. @jvillarf has requested to follow the same convention used in stats: `(undefined)` when no errors 
`({xxxxxxxxxxxxxxx})` when error exists

Issue #797, opened 2017-11-20
Optimize widgets (dataviews) queries


rafatower:
Optimize widgets (dataviews) queries for PG10 PARALLEL processing.

Acceptance: widgets E2E benchmark show a significant improvement.

cc/ @Algunenano @jgoizueta

rafatower:
Please start with formula (avg) and categories histogram.

Use the benchmarks as your measure of progress. cc/ @jgoizueta

Algunenano:
Starting with formulae:

Sample query:
```sql
SELECT
    avg(deptherror) AS result,
    (
        SELECT
            count(1)
        FROM
            (
            SELECT * FROM (SELECT * FROM all_month) _cdb_bbox_filter
            WHERE ST_Intersects(the_geom_webmercator, ST_Transform(ST_MakeEnvelope(-144.27246093750003,32.509761735919426,-106.78710937500001,66.37275500247458, 4326), 3857))
            ) _cdb_formula_nulls
        WHERE deptherror IS NULL
    ) AS nulls_count,
    (
       SELECT
            count(1)
        FROM (
            SELECT * FROM (SELECT * FROM all_month) _cdb_bbox_filter
            WHERE ST_Intersects(the_geom_webmercator, ST_Transform(ST_MakeEnvelope(-144.27246093750003,32.509761735919426,-106.78710937500001,66.37275500247458, 4326), 3857))
            ) __cdb_formula_infinities
        WHERE deptherror = 'infinity'::float OR deptherror = '-infinity'::float
    ) AS infinities_count,
    (
        SELECT
            count(1)
        FROM (
            SELECT * FROM (SELECT * FROM all_month) _cdb_bbox_filter
            WHERE ST_Intersects(the_geom_webmercator, ST_Transform(ST_MakeEnvelope(-144.27246093750003,32.509761735919426,-106.78710937500001,66.37275500247458, 4326), 3857))
            ) __cdb_formula_nans
        WHERE deptherror = 'NaN'::float
  ) AS nans_count

FROM (
    SELECT * FROM (SELECT * FROM all_month) _cdb_bbox_filter
    WHERE ST_Intersects(the_geom_webmercator, ST_Transform(ST_MakeEnvelope(-144.27246093750003,32.509761735919426,-106.78710937500001,66.37275500247458, 4326), 3857))
    ) __cdb_formula
[...]

PR #801, opened 2017-11-27
Formula widgets: Optimize and cleanup DB query

Algunenano:
Optimize formula queries.

Since the formula and histogram changes (#797) are going to be kind of large I've decided to break the changes in several PRs (not sure if 2 or 3 (numeric histogram, date histogram, formula)).

### DIFF
--- a/lib/cartodb/models/dataview/formula.js
+++ b/lib/cartodb/models/dataview/formula.js
@@ -1,34 +1,21 @@
 const BaseDataview = require('./base');
 const debug = require('debug')('windshaft:dataview:formula');
 
-const countInfinitiesQueryTpl = ctx => `
-    SELECT count(1) FROM (${ctx.query}) __cdb_formula_infinities
-    WHERE ${ctx.column} = 'infinity'::float OR ${ctx.column} = '-infinity'::float
-`;
-
-const countNansQueryTpl = ctx => `
-    SELECT count(1) FROM (${ctx.query}) __cdb_formula_nans
-    WHERE ${ctx.column} = 'NaN'::float
-`;
-
-const filterOutSpecialNumericValuesTpl = ctx => `
-    WHERE
-        ${ctx.column} != 'infinity'::float
-    AND
-        ${ctx.column} != '-infinity'::float
-    AND
-        ${ctx.column} != 'NaN'::float
-`;
-
+/* jshint ignore:start */
 const formulaQueryTpl = ctx => `
-    SELECT
-        ${ctx.operation}(${ctx.column}) AS result,
-        (SELECT count(1) FROM (${ctx.query}) _cdb_formula_nulls WHERE ${ctx.column} IS NULL) AS nulls_count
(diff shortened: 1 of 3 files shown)


lib/cartodb/models/dataview/formula.js
```diff
@@ -1,34 +1,21 @@
...
-const filterOutSpecialNumericValuesTpl = ctx => `
-    WHERE
-        ${ctx.column} != 'infinity'::float
-    AND
-        ${ctx.column} != '-infinity'::float
-    AND
-        ${ctx.column} != 'NaN'::float
-`;
-
+/* jshint ignore:start */
```
rafatower: This is kind of cheating and should be avoided. I guess jshint was complaining about something but don't know what that was. Cyclomatic complexity by any chance?

lib/cartodb/models/dataview/formula.js
```diff
@@ -1,34 +1,21 @@
...
 const formulaQueryTpl = ctx => `
-    SELECT
-        ${ctx.operation}(${ctx.column}) AS result,
-        (SELECT count(1) FROM (${ctx.query}) _cdb_formula_nulls WHERE ${ctx.column} IS NULL) AS nulls_count
-        ${ctx.isFloatColumn ? `,(${countInfinitiesQueryTpl(ctx)}) AS infinities_count` : ''}
-        ${ctx.isFloatColumn ? `,(${countNansQueryTpl(ctx)}) AS nans_count` : ''}
-    FROM (${ctx.query}) __cdb_formula
-    ${ctx.isFloatColumn && ctx.operation !== 'count' ? `${filterOutSpecialNumericValuesTpl(ctx)}` : ''}
-`;
+${!ctx.isFloatColumn ?
```
rafatower: I'd actually use 2 different templates

lib/cartodb/models/dataview/overviews/formula.js
```diff
@@ -6,50 +6,35 @@ var dot = require('dot');
 dot.templateSettings.strip = false;
 
 var formulaQueryTpls = {
+/* jshint ignore:start */
```
rafatower: same here

lib/cartodb/models/dataview/formula.js
Algunenano: No, long lines and I didn't want to mess up the SQL query structure to make the js linter happy.

lib/cartodb/models/dataview/formula.js
Algunenano: I'll see what I can do

lib/cartodb/models/dataview/formula.js
rafatower: I totally agree with you, better not to split into several lines for the sake of readability.

You can just add a comment to state that.

Algunenano:
When working in the histograms some stuff has come up that could be put in common, so I'll close this and  open a single PR with several commits.

PR #804, opened 2017-11-29
Widgets query optimizations

### LINKED ISSUE #792: Time series dataview buckets values "bug"
### Context

We have an inconsistency in how Windshaft groups timestamps into bins and how UI needs / show them. 

### Steps to Reproduce

1. Import this dataset in Builder (rename to .csv) 
[dospuntos copy.txt](https://github.com/CartoDB/Windshaft-cartodb/files/1470566/dospuntos.copy.txt)
2. Create a time series widget
3. Set the widget timezone to `Azores` (UTC - 1) and aggregation to days.

We have two points with timestamps:
1. Soria `1970-01-02 00:00:00+00` Timestamp 86400
2. Zaragoza `1970-01-02 01:00:00+00` Timestamp 90000

The API returns 2 buckets

```
{
  "aggregation": "day",
  "offset": -3600,
  "timestamp_start": 0,
  "bin_width": 1800,
  "bins_count": 2,
  "bins_start": 3600,
  "nulls": 0,
  "bins": [
    {
      "bin": 0,
      "min": 86400,
      "max": 86400,
      "avg": 86400,
      "freq": 1,
      "timestamp": 3600
    },
    {
      "bin": 1,
      "min": 90000,
      "max": 90000,
      "avg": 90000,
      "freq": 1,
      "timestamp": 90000
    }
  ],
  "type": "histogram"
}
```

As you can see in the payload we have this info:
- Buckets start at `timestamp_start: 0` That is 1970-01-01 00:00:00
- First bucket has 1 point with the timestamp 86400. (1970-01-02 00:00:00)
- Second bucket has 1 point with the timestamp 90000. (1970-01-02 01:00:00) 

So if buckets start at 0 (day 1 00:00:00) and it includes 86400 (day 2 00:00:00) we have a problem. It includes a whole day **and a second** 😱 . 86400 should belong to the next bucket.

### Current Result

bucketing bug

As you can see in the GIF, we draw the two buckets returned by the API but when we filter we send whether day 1 (from 00:00:00 to 23:59:59) or day 2 (from 00:00:00 to 23:59:59)

Soria points was returned in the first bucket but when filtering from 86400 (day 2 00:00:00) it's included.

Algunenano:
Optimizations for histogram (numeric and time-series), formula and formula with overview queries.

Should handle #797 and fix #792. I'll generate some comparisons and add them in #797

### DIFF
--- a/test/acceptance/dataviews/histogram.js
+++ b/test/acceptance/dataviews/histogram.js
@@ -256,6 +256,7 @@ describe('histogram-dataview for date column type', function() {
     }];
 
     dateHistogramsUseCases.forEach(function (test) {
+
         it('should create a date histogram aggregated in months (EDT) ' + test.desc, function (done) {
             var OFFSET_EDT_IN_MINUTES = -4 * 60; // EDT Eastern Daylight Time (GMT-4) in minutes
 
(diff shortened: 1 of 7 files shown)


test/acceptance/dataviews/histogram.js
```diff
@@ -332,23 +332,8 @@ describe('histogram-dataview for date column type', function() {
...
-                assert.ok(!err, err);
-                assert.equal(dataview.type, 'histogram');
-                assert.ok(dataview.bin_width > 0, 'Unexpected bin width: ' + dataview.bin_width);
-
-                done();
-            });
-        });
-
-        it('should return same histogram ' + test.desc, function (done) {
+        /* This test is not valid when using 'without offset' in a timezone != UTC */
```
rafatower: It was added because of this: https://github.com/CartoDB/Windshaft-cartodb/pull/742

instead of disabling, it should be maybe improved

rafatower:
Just for the record: there's a minor change in the (undocumented) API: in histogram bins, min/max/avg are now equal to timestamp (not used anyway and agreed upon)

test/acceptance/dataviews/histogram.js
Algunenano: I've added again the removed test increasing the range so it does what it expects to do and, for the disabled one, I've changed it so it only tests with timestamptz values (where this is guaranteed).

PR #806, opened 2017-11-30
Mapconfig aggregation

dgaubert:


lib/cartodb/models/aggregation/aggregation-templates.js
```diff
@@ -0,0 +1,102 @@
...
+                WHERE the_geom_webmercator && _cdb_params.bbox
+                GROUP BY _cdb_gx, _cdb_gy
+            )
+            SELECT
+              ST_SetSRID(ST_MakePoint(_cdb_gx*(res+0.5), _cdb_gy*(res*0.5)), 3857) AS the_geom_webmercator,
+              _cdb_feature_count
+              FROM _cdb_clusters, _cdb_params
+        `,
+
+        'point-sample-': ctx => `
```
rochoa: Is the trailing hyphen intended?

lib/cartodb/models/mapconfig/adapter/aggregation-mapconfig-adapter.js
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+        const aggregationMetadata = queryUtils.getAggregationMetadata({ query: layer.options.sql });
+
+        connection.query(aggregationMetadata, (err, res) => {
+            if (err) {
+                return callback(null, shouldAdapt);
+            }
+
+            const estimatedFeatureCount = res.rows[0].count;
```
rochoa: I would check `rows[0]` exists:
```
const result = res.rows[0] || {};
const estimatedFeatureCount = result.count || ...;
```

lib/cartodb/models/mapconfig/adapter/aggregation-mapconfig-adapter.js
```diff
@@ -0,0 +1,225 @@
...
+
+            const threshold = layer.options.aggregation && layer.options.aggregation.threshold ?
+                layer.options.aggregation.threshold :
+                1e5;
```
rochoa: I see this constant here and also in `aggregation-proxy.js`. I'm not sure about the number itself, but I would try to keep it just at one place.

lib/cartodb/models/mapconfig/adapter/aggregation-mapconfig-adapter.js
```diff
@@ -0,0 +1,225 @@
+const AggregationProxy = require('../../aggregation/aggregation-proxy');
+const { MapConfig } = require('windshaft').model;
+const queryUtils = require('../../../utils/query-utils');
+
+const MISSING_AGGREGATION_COLUMNS = 'Missing columns in the aggregation. The map-config defines cartocss expressions,'+
+' interactivity fields or attributes that are not present in the aggregation';
+const unsupportedGeometryTypeErrorMessage = ctx =>
+`Unsupported geometry type (${ctx.geometryType}) for aggregation. Aggregation is available only for points.`;
```
rochoa: I would be more explicit and say `ST_Point` as it is what we are using.

dgaubert:
Hey @rochoa !!

Thanks for reviewing. I would like to refactor `_hasAnyLayerAggregation()`, `_hasLayerAggregation()` to move them to `MapConfig` model in `Windshaft`. From my point of view `MapConfig` is the one that should know about its internals. 

What do you think?. It puts more pressure on `MapConfig` model and we will have to refactor it in order to have a family of different subclasses overriding methods of a common interface and so on...  (this is out of the scope of this PR, obviously..)

lib/cartodb/models/aggregation/aggregation-templates.js
dgaubert: @jgoizueta ^^

jgoizueta: It's a typo, should be `point-sample`

PR #808, opened 2017-12-11
Force png tile generation for static maps

ethervoid:
If the user tries to generate a static map requesting JPG it will fail
because is going to try to generate the tiles using JPG as format which
is not supported by now, this bug was introduced in the version 4.0.1

So we now force, again, the tiles to be generated as PNG but we pass
the requested format, JPG, to windshaft to generate the final image as
the user reqests


lib/cartodb/controllers/layergroup.js
```diff
@@ -357,7 +357,9 @@ LayergroupController.prototype.center = function(req, res, next) {
 
 LayergroupController.prototype.staticMap = function(req, res, width, height, zoom /* bounds */, center, next) {
     var format = req.params.format === 'jpg' ? 'jpeg' : 'png';
-    req.params.format = req.params.format || 'png';
```
ethervoid: @oleurud I think this was a bug too or am I mistaken? (this should be res.locals.format isn't it?)

oleurud: Yes, you are right, it was a bug

oleurud: :fearful:

ethervoid: I'm adding test for jpg too

ethervoid:
@oleurud test added plus some refactor to support different formats in the assertion

Issue #776, opened 2017-10-17
Changes to dataviews and filters


ivanmalagon:
As we continue the development of `cartodb.js v4` we realized that we need some changes on dataviews and filters.

Until now, dataviews acted as a proxy to filters. We want to split this, having filters and dataview separated to reflect better the underlying architecture of the system.

There is an ongoing PR on https://github.com/CartoDB/Windshaft-cartodb/compare/analyses-filters that enables to send the filter parameters attached to analysis nodes instead of dataviews.

#### Problem 1: own_filter

The problem is that because of these last changes, calling the dataview URL with `own_filter` has stopped working. It's time to rethink the endpoint.

After some research, we need three ways of calling a dataview endpoint:

1. Fetch data from analysis/source with No filters applied. 

2. Fetch data from analysis/source with All filters applied.

3. Fetch data from analysis/source with All filters except the one applied to the same analysis and column.

Number 3 is not working now. Prior to this, the filters had the same id as the dataview so it was easy to relate them. Now it seems that this link is broken. Debugging https://github.com/CartoDB/Windshaft-cartodb/blob/73a633ae7d19cdcfae1e45e20090483e37251e07/lib/cartodb/models/mapconfig/adapter/analysis-mapconfig-adapter.js#L142 we can see that now with last changes, `own_filter_on` and `own_filter_off` return the same query.

![image](https://user-images.githubusercontent.com/1078228/31660569-175f4e66-b338-11e7-9bc4-386a812cf2db.png)

So, we need:

- A parameter to specify what data we want to retrieve. It could be `filters` with 3 possible values:
    - `?filters=none`: no filters applied.
    - `?filters=all`: all filters applied.
    - `?filters=others`: all filters except the one applied to the same analysis and column. Default.

#### Problem 2: category search

In category dataview URL there is another options that it's used to search categories.

```
dataviewUrl/search?q=text
```

With the last changes, that search has stopped working since it's affected by any filter applied to the same analysis and column.

We need that endpoint to search in all categories with no filters applied.

alonsogarciapablo:
> ?filters=others: all filters except the one applied to the same analysis and column.

We have to take into account that multiple filters can be applied to the same analysis and column at the same time.

ivanmalagon:
There's an ongoing work in branch https://github.com/CartoDB/Windshaft-cartodb/tree/analyses-filters-params

@dgaubert Two things:
[...]

PR #809, opened 2017-12-12
Add `no_filters` param to dataviews

ivanmalagon:
Related to https://github.com/CartoDB/Windshaft-cartodb/issues/776
----

This PR adds the query param `no_filters` to the dataview URL in order to get the data with no filters applied.


dgaubert:
What if I request a dataview's result with `own_filter=1` and `no_filters=1`? 

If I'm not wrong, the current implementation returns a dataview without filters. Is that the expected behavior??

ivanmalagon:
If both params are present, `no_filters` prevails and `own_filter` is discarded.

For me, that's the expected behaviour since it's a stronger condition.

dgaubert:
Let me be a PITA.

I don't agree. I understand your point but it's confusing to the user's point of view. We have two different params (`no_filters` and `own_filter`) for the same thing: to choose the dataview's query. The fact of having two different params was for being more user-friendly, giving priority to them it's confusing.

I'd validate when both of them are present and respond with error:
```
if (Number.isFinite(ownFilter) && Number.isFinite(noFilters)) {
  const error = new Error('Bad query params ....');
  error.http_status = 400
  return callback(error);
}
```

ivanmalagon:
Can you take a look on the commit https://github.com/CartoDB/Windshaft-cartodb/pull/809/commits/02ac25181e58084937bc46509f640775c8e6e97f ?

I've added a `type='dataview'` to get a better context to tackle the error in Builder, but I haven't found any other appearance of it. So if there's a better type for the dataview let me know. (didn't fell the `unknown` type)

PR #814, opened 2017-12-15
Fix agg query for point-grid placement

rochoa:


rochoa:
A "FERNANDO IS FASTER THAN YOU" situation.

PR #815, opened 2017-12-18
Torque tests

Algunenano:
Couple of fixes related to tests with pg10:
- Order torque objects before comparison (since order isn't guaranteed by the query it was failing the comparisons).
- Disable timeout tests failing because mvt and torque use different pg clients so setting the timeout by parameters did not have an effect.

@jgoizueta Can you please have a look?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@ Released 2018-01-04
 New features:
  - Return url template in metadata #838.
 
+Bux fixes:
+ - Tests: Order torque objects before comparison
 
 ## 4.7.0
 Released 2018-01-03
@@ -29,7 +31,6 @@ Announcements:
  - Validate aggregation input params.
  - Fix column names collisions in histograms [#828](https://github.com/CartoDB/Windshaft-cartodb/pull/828).
  - Add full-sample aggregation support for vector map-config.
-
 
 ## 4.5.0
 Released 2017-12-19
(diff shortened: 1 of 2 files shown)


test/acceptance/user-database-timeout-limit.js
```diff
@@ -625,7 +625,7 @@ describe('user database timeout limit', function () {
                     this.testClient.setUserDatabaseTimeoutLimit(0, done);
                 });
 
-                it('"torque.json" fails due to statement timeout', function (done) {
+                xit('"torque.json" fails due to statement timeout', function (done) { // jshint ignore:line
```
dgaubert: We cannot disable these tests. They are acceptance ones and validates that limits v1 is working propertly.

In `beforeEach` hook, we are setting a timeout to the database and then refresing every connection in the pool which is the same for both torque and mvt.

BTW, to avoid `jshint` lints you can use `it.skip` instead of `xit` function.

dgaubert: If for some reason these tests fail with Postgres 10 we must find the root cause.

Algunenano: I'll remove that commit. I'm not sure if it fails due to PG10 or due to `node-pg` since I cannot use our current release (https://github.com/CartoDB/node-cartodb-psql/issues/30) in my system and I'm using a different one (github:cartodb/node-postgres#cdb-6.1)

Algunenano:
Hey @dgaubert Is this ok to be merged with just the fix for torque tests?

PR #817, opened 2017-12-18
Date histogram: Add second, decade, century and millennium aggregations

Algunenano:
Add support for aggregations by second, decade, century or millennium.

I've tested it with Builder and it doesn't break anything, that is returning a new aggregation when auto is requested does not cause any apparent issue and it's reused correctly. For example if it returns "second" the client will request "second" subsequently.

Also I've reduced the number of max bins in auto from 366 to 100 to give more priority to bigger aggregations. Keep in mind that this only has an effect when using "auto", it doesn't have any impact if a specific aggregation is requested.

### DIFF
--- a/lib/cartodb/models/dataview/histograms/date-histogram.js
+++ b/lib/cartodb/models/dataview/histograms/date-histogram.js
@@ -141,8 +141,8 @@ const dateIntervalQueryTpl = ctx => `
     FROM __cdb_interval_in_days, __cdb_interval_in_hours, __cdb_interval_in_minutes, __cdb_interval_in_seconds
 `;
 
-
-const MAX_INTERVAL_VALUE = 366;
+/** Constant to switch between aggregations in auto mode */
+const MAX_INTERVAL_VALUE = 100;
 
 const DATE_AGGREGATIONS = {
     'auto': true,
@@ -250,9 +250,9 @@ ORDER BY bin ASC;
                 return callback(err);
             }
 
-            const aggegations = result.rows[0];
-            const aggregation = Object.keys(aggegations)
-                .map(key => ({ name: key, value: aggegations[key] }))
+            const aggregations = result.rows[0];
+            const aggregation = Object.keys(aggregations)
+                .map(key => ({ name: key, value: aggregations[key] }))
                 .reduce((closer, current) => {
                     if (current.value > MAX_INTERVAL_VALUE) {
                         return closer;
--- a/lib/cartodb/models/dataview/histograms/date-histogram.js
+++ b/lib/cartodb/models/dataview/histograms/date-histogram.js
@@ -127,8 +127,11 @@ const dateIntervalQueryTpl = ctx => `
         FROM __cdb_interval_in_minutes, __cdb_dates
     )
     SELECT
+        ROUND(__cdb_days / 365243) AS millennium,
+        ROUND(__cdb_days / 36525) AS century,
+        ROUND(__cdb_days / 3652) AS decade,
         ROUND(__cdb_days / 365) AS year,
-        ROUND(__cdb_days / 90) AS quarter,
+        ROUND(__cdb_days / 91) AS quarter,
         ROUND(__cdb_days / 30) AS month,
         ROUND(__cdb_days / 7) AS week,
         __cdb_days AS day,
@@ -143,13 +146,17 @@ const MAX_INTERVAL_VALUE = 366;
 
 const DATE_AGGREGATIONS = {
     'auto': true,
+    'second' : true,
     'minute': true,
     'hour': true,
     'day': true,
     'week': true,
     'month': true,
     'quarter': true,
-    'year': true
+    'year': true,
+    'decade' : true,
+    'century' : true,
+    'millennium' : true
 };
 
 /**
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 Released 2017-mm-dd
 
 Announcements:
+ - Date histograms: Add second, decade, century and millenium aggregations
+ - Date histograms: Switch the auto threshold from 366 buckets to 100.
  - Logging all errors.
  - Add support for aggregated visualizations.
  - Allow vector-only map-config creation.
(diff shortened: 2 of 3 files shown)


lib/cartodb/models/dataview/histograms/date-histogram.js
```diff
@@ -127,8 +127,11 @@ const dateIntervalQueryTpl = ctx => `
         FROM __cdb_interval_in_minutes, __cdb_dates
     )
     SELECT
+        ROUND(__cdb_days / 365243) AS millennium,
+        ROUND(__cdb_days / 36525) AS century,
+        ROUND(__cdb_days / 3652) AS decade,
         ROUND(__cdb_days / 365) AS year,
-        ROUND(__cdb_days / 90) AS quarter,
+        ROUND(__cdb_days / 91) AS quarter,
```
dgaubert: Just curious, why does 91 fit better than 90?

lib/cartodb/models/dataview/histograms/date-histogram.js
```diff
@@ -138,18 +141,22 @@ const dateIntervalQueryTpl = ctx => `
     FROM __cdb_interval_in_days, __cdb_interval_in_hours, __cdb_interval_in_minutes, __cdb_interval_in_seconds
 `;
 
-
-const MAX_INTERVAL_VALUE = 366;
+/** Constant to switch between aggregations in auto mode */
+const MAX_INTERVAL_VALUE = 100;
```
dgaubert: This will change the behavior in builder for current time-series in Builder. Did you talk with anyone from @CartoDB/frontend?

lib/cartodb/models/dataview/histograms/date-histogram.js
```diff
@@ -138,18 +141,22 @@ const dateIntervalQueryTpl = ctx => `
...
     'auto': true,
+    'second' : true,
     'minute': true,
     'hour': true,
     'day': true,
     'week': true,
     'month': true,
     'quarter': true,
-    'year': true
+    'year': true,
```
dgaubert: it might that Builder receives an aggregation that they don't know how to handle it. If I recall correctly they have to know what aggregation is being used in order to build the widget properly (to build widget's labels 🤔 ).

lib/cartodb/models/dataview/histograms/date-histogram.js
Algunenano: Not yet, I was planning to tell @ivanmalagon  as this is part of https://github.com/CartoDB/support/issues/1071. Keep in mind that this only affects which aggregation is selected when using "auto".

lib/cartodb/models/dataview/histograms/date-histogram.js
Algunenano: I tested it locally and it worked fine (received "auto", sent "seconds" and the next request used "seconds"; filters using the buckets worked fine too (1 second is this case). There still are some things to be implemented (UI and the likes) but I didn't see anything broken.

lib/cartodb/models/dataview/histograms/date-histogram.js
Algunenano: 365/4 = 91.25

dgaubert: haha! 
My approach was `30*3`. Fair enough! :)

lib/cartodb/models/dataview/histograms/date-histogram.js
dgaubert: I do believe that they use `auto`mode always.

lib/cartodb/models/dataview/histograms/date-histogram.js
dgaubert: if someone from @CartoDB/frontend can shed some light here it'd be much appreciated!

dgaubert: Well, do we return aggregation as `string` or as `seconds`?

lib/cartodb/models/dataview/histograms/date-histogram.js
Algunenano: Yes, it's used by default but you can change it

lib/cartodb/models/dataview/histograms/date-histogram.js
Algunenano: As a string, for example:
First request (uses `auto`):
```
http://dev.localhost:8181/api/v1/map/81d1bbb8b5455add281b45ff491c1c29:1513601345660/dataview/213e5193-cee8-4b04-a500-c4370f904111?aggregation=auto&api_key=cfd51af386ec59859a1415d51f9b1809c014d369
```
Response:
```json
{  
   "aggregation":"second",
   "offset":0,
   "timestamp_start":86400,
   "bin_width":1,
   "bins_count":6,
   "bins_start":86400,
   "nulls":0,
   "bins":[  
      {  
         "bin":0,
         "min":86400,
         "max":86400,
         "avg":86400,
         "freq":1,
         "timestamp":86400
      },
      {  
         "bin":5,
         "min":86405,
         "max":86405,
         "avg":86405,
         "freq":1,
         "timestamp":86405
      }
   ],
   "type":"histogram"
}
```

Next request (uses `second`):
```
http://dev.localhost:8181/api/v1/map/81d1bbb8b5455add281b45ff491c1c29:1513601345660/dataview/213e5193-cee8-4b04-a500-c4370f904111?bbox=-3.9605712890625004,40.455307212131494,0.5548095703125001,42.549033612225145&aggregation=second&offset=0&start=86400&end=86405&api_key=cfd51af386ec59859a1415d51f9b1809c014d369
```

Same response.

Algunenano: I've talked to @ivanmalagon and he's ready to pick it up in the front-end side so merging and releasing

Jesus89: I have just created a PR to Carto.js replacing `moment` by native calculations: https://github.com/CartoDB/cartodb.js/pull/1981.

Now it is very easy to add the new aggregations: https://github.com/CartoDB/cartodb.js/pull/1981/files#diff-3833e4ba62a3bd322d40631559fa4020R11

PR #821, opened 2017-12-18
Experimental full-sample aggregation

jgoizueta:
A new kind of _placement_ has been added, `full-sample` which takes a full record from each aggregated group. The count (`_cdb_features_count`) and any aggregation explicitely requested is also present in the result.


jgoizueta:
About the [last commit](https://github.com/CartoDB/Windshaft-cartodb/pull/821/commits/e9a4fc4b2c1188e3262a820de95ac38e31e36d4b):

The "full-sample" aggregation is now used only for the default case, when the aggregation parameters have no placement, columns or dimensions.

This special mode performs only spatial aggregation (using a grid defined by the requested tile and the resolution, parameter, as all the other cases),
and returns a _random_ record from each group (grid cell) with all its columns and an additional `_cdb_features_count` with the number of features in the group.

Regarding the randomness of the sample: currently we use the row with the minimum `cartodb_id` value in each group.

When either a explicit placement, columns or dimensions are requested we no longer use the special, query; we use one determined by the placement (which will default to "centroid"),
and it will have as columns only the aggregated columns and dimensions specified, in addition to `_cdb_features_count`, which is always present.

We might decide in the future to allow sampling column values for any of the different placement modes, but at the moment that

The rationale behind having this special aggregation with all the original columns is to provide a mostly transparent way to handle large datasets without having to provide special map configurations for those cases.
(i.e. preserving the logic used to produce the maps with smaller datasets). Overviews have been used so far with this intent, but they are inflexible.

dgaubert:
Superseded by https://github.com/CartoDB/Windshaft-cartodb/pull/830/

PR #834, opened 2018-01-02
Middlewarify named maps controller

dgaubert:


lib/cartodb/controllers/named_maps.js
```diff
@@ -217,118 +143,223 @@ NamedMapsController.prototype.prepareLayerFilterFromPreviewLayers = function (
...
+            res.locals.body = image;
+            res.locals.headers = headers;
+            res.locals.stats = stats;
+
+            next();
+        });
+    }.bind(this);
+};
+
+const incrementMapViewsError = ctx => `ERROR: failed to increment mapview count for user '${ctx.user}': ${ctx.err}`;
```
enekid: I would use an old fashioned function instead of those fancy arrow functions (I've watched some Kyle Simpson videos these holidays....)

```javascript
function incrementMapViewsError (ctx) {
 return `ERROR: failed to increment mapview count for user '${ctx.user}': ${ctx.err}`;
}
```

lib/cartodb/controllers/named_maps.js
```diff
@@ -339,9 +370,10 @@ function templateZoomCenter(view) {
 
 function templateBounds(view) {
     if (view.bounds) {
-        var hasAllBounds = _.every(['west', 'south', 'east', 'north'], function(prop) {
+        var hasAllBounds = ['west', 'south', 'east', 'north'].every(function(prop) {
```
enekid: IMHO in this case a arrow function seems better.

```javascript
var hasAllBounds = ['west', 'south', 'east', 'north'].every(bound => Number.isFinite(view.bounds[bound]));
```

lib/cartodb/controllers/named_maps.js
dgaubert: Hey, thanks for the advice!

Could you explain it to me a little? I've been using arrow function to create template strings for months, it might I must stop using them for this use case.

lib/cartodb/controllers/named_maps.js
dgaubert: 👍

lib/cartodb/controllers/named_maps.js
enekid: In this case we are defining a function global to the file. We don't benefit from the use of an arrow function (```const xxx = () => ...```). The letter count is not much less and we don't make use of ```this```.

IMHO in such cases using an arrow function:

- we loose clarity (no function keyword) 
- more difficult to debug (anonymous function)

More examples [When 'not' to use arrow functions](https://dmitripavlutin.com/when-not-to-use-arrow-functions-in-javascript/)

I would use arrow functions only in map/reduce type of calls and when a special ```this``` handling is needed.

dgaubert: I couldn't be more agree. I've started to use them because they are so fancy:

```js
const template = ctx => `I'm not really cool, ${ctx.name}`
```

So, I stop to use them in this kind of scenario.

thanks.

PR #851, opened 2018-01-18
use original column name in group by instead of alias

enekid:
There is some problem using in the `GROUP BY` an aliased property defined in the `SELECT`:

```sql
   SELECT
        CAST(company_na AS text) AS category,
        count(company_na) AS value
    FROM scoutarticles
    GROUP BY
       category
```

There're some solutions:
1. use in the `GROUP BY` the index of the property in the `SELECT`:
```sql
   SELECT
        CAST(company_na AS text) AS category,
        count(company_na) AS value
    FROM scoutarticles
    GROUP BY
       1
```

2. use in the `GROUP BY` the original column name:
```sql
   SELECT
        CAST(company_na AS text) AS category,
        count(company_na) AS value
    FROM scoutarticles
    GROUP BY
       company_na
```

I've implemented option 2.


dgaubert:
Hey! 

Where is the automatic test? I can't see it.

PR #854, opened 2018-01-26
Respect category type of aggregation dataview 

oleurud:
Currently, we cast the `category` field to string, regardless of what real type of the `category` field.

So, if we aggregate a numeric field, we are returning:
```
...
"categories": [
    {
      "category": "10",
      "value": 1,
      "agg": false
    },
    {
      "category": "4",
      "value": 1,
      "agg": false
    },
...
```

This behavior is a problem for the frontend https://github.com/CartoDB/support/issues/1248 that need to receive the real type of the value.

### DIFF
--- a/lib/cartodb/models/dataview/aggregation.js
+++ b/lib/cartodb/models/dataview/aggregation.js
@@ -74,7 +74,7 @@ const specialNumericValuesColumns = () => `, nans_count, infinities_count`;
 
 const rankedAggregationQueryTpl = ctx => `
     SELECT
-        CAST(category AS text),
+        category,
         value,
         false as agg,
         nulls_count,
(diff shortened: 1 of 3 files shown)


lib/cartodb/models/dataview/aggregation.js
```diff
@@ -292,7 +292,13 @@ module.exports = class Aggregation extends BaseDataview {
             min: min_val,
             max: max_val,
             categoriesCount: categories_count,
-            categories: result.rows.map(({ category, value, agg }) => ({ category, value, agg }))
+            categories: result.rows.map(({ category, value, agg }) => {
```
enekid: This could be rewritten in ES6 as:

```javascript
categories: result.rows.map({ category, value, agg } => 
              { 
                category: agg ? 'Other' : category,
                value, 
                agg 
              };
           )
```

But don't change it if you find it clearer with your version.

PR #859, opened 2018-01-31
Add Mapnik 'cache-features' option

Algunenano:


rafatower:
> Note this would require changing the configuration in the production machines in order to actually enable the cache-features setting.

Actually, as Raul pointed out to me, the behavior in Windshaft is to have `cache-features` enabled by default. So a config change is not required but could be disabled, if ever needed.

rafatower:
PS: you'll need to `yarn install` and then commit and push the `yarn.lock` as there are changes in the dependencies.

Algunenano:
Added a commit to update `yarn.lock`.

Algunenano:
Switched to `5.3.3` instead so it doesn't allow setting the option though Mapconfig.

PR #862, opened 2018-02-02
[DMY] dont calculate webmercator in merge node

oleurud:
In the JOIN analysis (aka merge node), we are calculating the `the_geom_webmercator` on the fly. This analysis is not cached, so it is not necessary to calculate it because the source table already has the `the_geom_webmercator` [more info](https://github.com/CartoDB/camshaft/issues/200#issuecomment-296974271)

Do first this camshaft PR --->  https://github.com/CartoDB/camshaft/pull/349

Both PRs fix https://github.com/CartoDB/support/issues/1034 but not the related with `couldn't project point (180 -90 0)` PostGIS error

### DIFF
--- a/lib/cartodb/models/mapconfig/adapter/analysis-mapconfig-adapter.js
+++ b/lib/cartodb/models/mapconfig/adapter/analysis-mapconfig-adapter.js
@@ -184,8 +184,17 @@ var wrappedQueryTpl = dot.template([
     'FROM ({{=it._query}}) _cdb_analysis_query'
 ].join('\n'));
 
+// determine if a node has a usable the_geom_webmercator column;
+// otherwise to obtain it we'll need to ST_Tranform the_geom.
+function useTheGeomWebmercator(node) {
+    // * source nodes should always have an indexed the_geom_webmercator, so use it
+    // * cached nodes may have the_geom_webmercator, but they have an index on `ST_Transform(the_geom, 3857)`,
+    //   so we better use that expression instead.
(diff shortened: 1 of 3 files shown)


lib/cartodb/models/mapconfig/adapter/analysis-mapconfig-adapter.js
```diff
@@ -185,7 +185,7 @@ var wrappedQueryTpl = dot.template([
 ].join('\n'));
 
 function layerQuery(node) {
-    if (node.type === 'source') {
+    if (node.type === 'source' || node.type === 'merge') {
```
jgoizueta: It would be better to be able to check the output columns of the node.

oleurud: Yes, you are right

jgoizueta:
I'm closing this because:
* While writing analysis nodes it has not been a requirement to keep consistency of the_geom_webmercator with the_geom. The *official* geometry is the_geom, so an analysis could by accident include the_geom_webmercator from one of its inputs which does not correspond to the_geom.
* As it happens all current analyses that have the_geom_webmercator do it right and it corresponds to the_geom. But this is not easy to check/enforce for future development
* Cached analyses may have the_geom_webmercator, but it's not indexed. They do have an index on `ST_Transform(the_geom, 3857)`. So it's more efficient to use the latter (even though it may fail if conversion is not possible).

We'll try to fix the case we intended to solve here (failing conversions) by using some efficient and safe alternative to `ST_Transform`.

PR #863, opened 2018-02-02
removing raster image on timeout error

oleurud:
Tile render timeout is showing cartofante and we don't want to show it anymore.

The real change to remove it is an infrastructure task (changing the config example file, nothing happens), but I create this PR for consistency

More info --> https://github.com/CartoDB/cartodb/issues/13447


dgaubert:
Do we have news about these changes being promoted to production?

oleurud:
These changes don't affect the app. So, we can deploy whenever.

The infra team did the change some time ago --> https://github.com/CartoDB/cartodb-platform/pull/3951/files

PR #865, opened 2018-02-07
Improve the speed of the aggregation dataview

rafatower:
Improve the performance of the aggregation dataview.

Instead of using a CTE (WITH) for filtered_source, which is only used in
one place to calculate ranks, inject it as a subquery.

This way the planner has a chance to ignore uneeded columns as well as
to parallelize the exectution of the window function (WindowAgg in the
query plan).

That is the part that takes most of the time of the query.

The improvement is about 30-40% in speed on PG10 with 4 cores. With PG9.5 it is slightly better, but by a tiny 1%, depending on the underlying table and specific request.

### DIFF
--- a/lib/cartodb/models/dataview/aggregation.js
+++ b/lib/cartodb/models/dataview/aggregation.js
@@ -2,19 +2,17 @@ const BaseDataview = require('./base');
 const debug = require('debug')('windshaft:dataview:aggregation');
 
 const filteredQueryTpl = ctx => `
-    (
-        SELECT *
-        FROM (${ctx.query}) _cdb_filtered_source
-        ${ctx.aggregationColumn && ctx.isFloatColumn ? `
-        WHERE
-            ${ctx.aggregationColumn} != 'infinity'::float
-        AND
-            ${ctx.aggregationColumn} != '-infinity'::float
-        AND
-            ${ctx.aggregationColumn} != 'NaN'::float` :
-        ''
-        }
-    ) filtered_source
+    SELECT *
+    FROM (${ctx.query}) _cdb_filtered_source
+    ${ctx.aggregationColumn && ctx.isFloatColumn ? `
+    WHERE
+        ${ctx.aggregationColumn} != 'infinity'::float
+    AND
+        ${ctx.aggregationColumn} != '-infinity'::float
+    AND
+        ${ctx.aggregationColumn} != 'NaN'::float` :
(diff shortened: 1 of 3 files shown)


lib/cartodb/models/dataview/aggregation.js
```diff
@@ -43,7 +43,7 @@ const rankedCategoriesQueryTpl = ctx => `
             ${ctx.column} AS category,
             ${ctx.aggregationFn} AS value,
             row_number() OVER (ORDER BY ${ctx.aggregationFn} desc) as rank
-        FROM filtered_source
+        FROM ${filteredQueryTpl(ctx)}
```
dgaubert: What about of doing:
```
FROM (${filteredQueryTpl(ctx)}) filtered_source
```
and filtered query template definition would look like:
```
const filteredQueryTpl = ctx => `
  SELECT *
   [..]
`;
```
it might we can use it in other places.

rafatower: +1 to this, for the sake of clarity

I tested other replacements of CTE by subqueries but in practice turned out to be slower.

Algunenano:
I need to have a deeper look into the exact query sent to Postgres to see if more improvements are possible, but I'd switch the summary query to look something like the one in [the formula widget](https://github.com/CartoDB/Windshaft-cartodb/blob/aa57cdefb370ef577bad6dacf26b1a88d82feb3a/lib/cartodb/models/dataview/formula.js#L6) which handles floats faster (at uses aux functions).

Algunenano:
If you have a recommended datasets that uses dataviews I'll test it in my main machine tomorrow and see if I can come up with any other improvements once I see the query and the plan, but the change in the PR is good and could be merged right meow.

rafatower:
I'd go for deploying this and creating specific tickets for the dataviews and widgets that we can still optimize instead.

Algunenano:
This was released as part of `5.3.1`.

Testing with local machine:
* Postgresql: 11devel
* Dataset: yellow_1m (1 million rows from one of the NY Taxi datasets).
* Widget covering most of the points.
* Timings before this change (warn DB):
```
1072.165
1015.840
1079.679
1052.620
1046.132
1017.587
```
* Timings after this change (warn DB):
```
741.660
755.639
758.388
710.836
757.422
713.109
```

All in all, great improvement.

PR #866, opened 2018-02-08
Update camshaft to 0.61.2

jgoizueta:
This is to fixe a bug in the line-sequential analyses (length was incorrect).

See https://github.com/CartoDB/camshaft/pull/351


Algunenano:
NEWS please :smile_cat:

PR #874, opened 2018-02-14
add fallback for using metadata fallback

enekid:
needs:

```
cartodb/node-cartodb-redis#project-auth-api-fallback
```

A test still fails: _can instanciate a template by id wadus_

### DIFF
--- a/lib/cartodb/api/auth_api.js
+++ b/lib/cartodb/api/auth_api.js
@@ -70,7 +70,7 @@ AuthApi.prototype.authorizedByAPIKey = function(user, res, callback) {
 
     this.metadataBackend.getApikey(user, apikeyToken, (err, apikey) => {
         if (err) {
-            if (err.message && -1 !== err.message.indexOf('name not found')) {
+            if (isNameNotFoundError(err)) {
                 err.http_status = 404;
             }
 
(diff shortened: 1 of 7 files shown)


lib/cartodb/api/auth_api.js
```diff
@@ -70,7 +72,22 @@ AuthApi.prototype.authorizedByAPIKey = function(user, res, callback) {
...
+        //Remove this block when Auth fallback is not used anymore
+        // AUTH_FALLBACK
+        if (!apikey.databasePassword && global.environment.postgres.password) {
+            apikey.databasePassword = global.environment.postgres.password;
+        }
+
         if ( !isValidApiKey(apikey)) {
+            return callback(null, true); // AUTH_FALLBACK :S If api key not found, use default_public
+
             const error = new Error('Unauthorized');
```
dgaubert: 😱 dead code?. it should be commented, shouldn't it?

enekid: It's ugly as hell, I know. When we remove the Auth Fallback, we just need to remove all the lines/blocks marked with `AUTH_FALLBACK`. 

However, this line could probably be removed, because in the `node-cartodb-redis` fallback I assure at least the default api key is returned: 

https://github.com/CartoDB/node-cartodb-redis/blob/e80e3042d56a4e55fd1eb0207847844394764e04/lib/carto_metadata.js#L647

enekid: Line removed!

Issue #820, opened 2017-12-18
Rate limit middleware


oleurud:
**Goals**:
- [x] Ensure endpoints groups (for now, only working in a route)
- [x] get rate limit
- [x] LUA script
- [x] response header or error
- [x] Check if possible using EVALSHA command to remember LUA script
- [x] Configuration parameters to enable/disable all or one endpoint group rate limits
- [x] Travis (Dockerfile and tests)

**To consider**:
- We must use a new Redis db for this purpose

oleurud:
Initially, we have had doubts about remaining value https://github.com/brandur/redis-cell/issues/12 of the `redis-cell`, but, after some research, @Tonino have created a PR https://github.com/brandur/redis-cell/pull/13 solving the problem.

PR #875, opened 2018-02-15
Rate limit middleware

oleurud:
Fix https://github.com/CartoDB/Windshaft-cartodb/issues/820
Requires https://github.com/CartoDB/node-cartodb-redis/pull/22


lib/cartodb/middleware/rate-limit.js
```diff
@@ -0,0 +1,190 @@
+'use strict';
+
+const REDIS_DB = 8;
```
dgaubert: 8? I thought that Redis had only five databases, from the docs I read it has 16 databases (from 0 to 15) 😱 . BTW, is that agreed with backend team? Why not use database 5?

lib/cartodb/middleware/rate-limit.js
```diff
@@ -0,0 +1,190 @@
...
+
+const RATE_LIMIT_ENDPOINTS_GROUPS = {
+    ENDPOINT_1: 'anonymous',
```
dgaubert: Why do you use numbers instead of a meaningful names? For instance, `ANONYMOUS_ENDPOINTS: 'anonymous'`

lib/cartodb/middleware/rate-limit.js
```diff
@@ -0,0 +1,190 @@
...
+    ENDPOINT_11: 'named_create',
+    ENDPOINT_12: 'named_get',
+    ENDPOINT_13: 'named',
+    ENDPOINT_14: 'named_update',
+    ENDPOINT_15: 'named_delete',
+    ENDPOINT_17: 'named_tiles'
+};
+
+function rateLimitMiddleware(metadataBackend, endpointGroup = null) {
+    return function rateLimit(req, res, next) {
```
dgaubert: I've been using Rochoa's convention that is just the contrary ;) 

```
function rateLimit (metadataBackend, endpointGroup = null) {
     return function rateLimitMiddleware (req, res, next) {
[...]
```
I don't mind to use either Rochoa's one or yours, but I'd like to be as predictable as possible.

lib/cartodb/middleware/rate-limit.js
```diff
@@ -0,0 +1,190 @@
...
+        if (!global.environment.enabledFeatures.rateLimitsEnabled) {
```
dgaubert: A micro-optimization: Why not check `rateLimitsEnabled` while composing middleware stack in route registration?? Thus we avoid checking this every time.

lib/cartodb/middleware/rate-limit.js
```diff
@@ -0,0 +1,190 @@
...
+function validRatelimit(rateLimit) {
+    return rateLimit.length === 5;
+}
+
+function isRateLimitEnabledByEndpoint(endpointGroup) {
+    return global.environment.enabledFeatures.rateLimitsByEndpoint[endpointGroup] === true;
+}
+
+
+module.exports.rateLimitMiddleware = rateLimitMiddleware;
```
dgaubert: When I see namespaced exports I tend to agree, but this module is a middleware (because it's placed in `middleware` folder) so I'd export the middleware by default and apply namespaces to others. In fact, in ES6 you have default and namespaced exports:

```
import rateLimit, { RATE_LIMIT_ENDPOINTS_GROUPS } from `middlewares/rate-limit`
```
Wha do you think?

lib/cartodb/middleware/rate-limit.js
oleurud: Talking with backend, we finally decided to use a different Redis database from the usual to allow (easier at least) a future change of the Redis server, for example, if we want to control all the limits in a unique and external service.

lib/cartodb/middleware/rate-limit.js
oleurud: I want to do it :)

lib/cartodb/middleware/rate-limit.js
```diff
@@ -0,0 +1,190 @@
...
+                return next(err);
+            }
+
+            const rateLimit = getLowerRateLimit(rateLimits);
+
+            if (!rateLimit) {
+                return next();
+            }
+
+            const isBloqued = rateLimit[0];
```
dgaubert: you meant `isBlocked`?

lib/cartodb/middleware/rate-limit.js
```diff
@@ -0,0 +1,190 @@
...
+     /*jshint maxcomplexity:10 */
+    if (!ratelimits || !Array.isArray(ratelimits) || !ratelimits.length) {
+        return;
+    }
+
+    let minIndex = 0;
+    let minValue;
+    let currentIndex = 0;
+    for (let ratelimit of ratelimits) {
+        if (!validRatelimit(ratelimit)) {
```
dgaubert: It'd be really helpful to give a name to `ratelimit[0]` and `ratelimit[2]` to make it more readable: `const [ isBlocked, , remaining ] = rateLimit`

lib/cartodb/middleware/rate-limit.js
```diff
@@ -0,0 +1,190 @@
...
+            continue;
+        }
+
+        if (ratelimit[0] === 1) {
+            // rate limited
+            minIndex = currentIndex;
+            break;
+        }
+        
+        if (minValue === undefined || ratelimit[2] < minValue) {
```
dgaubert: `minValue` -> `minRemainingValue`??

lib/cartodb/middleware/rate-limit.js
```diff
@@ -0,0 +1,190 @@
...
+            // smaller remaining
```
dgaubert: with better names, you can remove comments :)

lib/cartodb/middleware/rate-limit.js
```diff
@@ -0,0 +1,190 @@
...
+    let minValue;
+    let currentIndex = 0;
+    for (let ratelimit of ratelimits) {
+        if (!validRatelimit(ratelimit)) {
+            currentIndex++;
+            continue;
+        }
+
+        if (ratelimit[0] === 1) {
+            // rate limited
```
dgaubert: Same here, the 90% of the comments in code can be avoided by giving really meaningful names to `variables` and `functions`

.travis.yml
```diff
@@ -5,10 +5,10 @@ services:
   - docker
 
 before_install:
-  - docker pull cartoimages/windshaft-testing
+  - docker pull oleurud/windshaft-carto
```
dgaubert: that's temporary, right?

oleurud: Yes, I need to talk with infra about the best place to have the redis cell library. 

Really, with this commit https://github.com/CartoDB/Windshaft-cartodb/pull/875/commits/a97e628520e59986bec237e13d8692227f5e9f05 we don't need a new version of the docker image. I will remove this change

lib/cartodb/middleware/rate-limit.js
oleurud: Yes, I like it. 

Do you like in this way?
```
function rateLimit(metadataBackend, endpointGroup = null) {
    if (!global.environment.enabledFeatures.rateLimitsEnabled) {
        return (req, res, next) => { next(); };
    }

    return function rateLimitMiddleware(req, res, next) {
```

I don't see a good way on the controllers

lib/cartodb/middleware/rate-limit.js
oleurud: I have changed it

lib/cartodb/middleware/rate-limit.js
oleurud: Ouch!

lib/cartodb/middleware/rate-limit.js
enekid: Without it serving as precedent, I prefer Rochoa's way. The first/wrapper function returns the actual middleware.

oleurud: :+1:

lib/cartodb/middleware/rate-limit.js
oleurud: :+1:

lib/cartodb/middleware/rate-limit.js
oleurud: :+1:

lib/cartodb/middleware/rate-limit.js
oleurud: Yes, I was not very inspired

lib/cartodb/middleware/rate-limit.js
oleurud: :+1:

lib/cartodb/middleware/rate-limit.js
oleurud: I have had to use `rateLimitFn`
```
function rateLimitFn(metadataBackend, endpointGroup = null) {
    return function rateLimitMiddleware(req, res, next) {
```
to avoid names collision

oleurud:
All features finished, ready to final CR.

Before merge it, we will need to make some changes: 
- change the test docker image --> we need to talk with infra about the best place to have the redis cell library and update it https://github.com/CartoDB/Windshaft/blob/rateLimits/Dockerfile
- config parameters: we need to update the config params on cartodb-platform (example https://github.com/CartoDB/cartodb-platform/pull/3951/files)

lib/cartodb/api/user_limits_api.js
```diff
@@ -146,7 +146,7 @@ UserLimitsApi.prototype.getRateLimit = function (user, endpointGroup, callback)
  * @param {string} endpointGroup 
  */
 function getStoreKey(user, endpointGroup) {
-    return `limits:rate:store:${user}:${endpointGroup}`;
+    return `limits:rate:store:${user}:maps:${endpointGroup}`;
```
enekid: 👍

lib/cartodb/api/user_limits_api.js
```diff
@@ -77,3 +87,145 @@ UserLimitsApi.prototype.getTimeoutRenderLimit = function (username, apiKey, call
...
+
+UserLimitsApi.prototype.preprareRateLimit = function () {
+    var self = this;
+
+    if (this.options.limits.rateLimitsEnabled) {
+        this.metadataBackend.redisCmd(
+            RATE_LIMIT_REDIS_DB, 
+            'SCRIPT', 
+            ['LOAD', getRateLimitLuaScript()], 
+            function (err, sha) {
```
enekid: what happens if there is an error?

oleurud: Nothing. It should work using `EVALSHA`, else it will use `EVAL`

lib/cartodb/middleware/rate-limit.js
```diff
@@ -0,0 +1,69 @@
...
+    NAMED_GET: 'named_get',
+    NAMED: 'named',
+    NAMED_UPDATE: 'named_update',
+    NAMED_DELETE: 'named_delete',
+    NAMED_TILES: 'named_tiles'
+};
+
+function rateLimitFn(userLimitsApi, endpointGroup = null) {
+    return function rateLimitMiddleware(req, res, next) {
+        if (!global.environment.enabledFeatures.rateLimitsEnabled) {
```
dgaubert: I prefer to return a `noop` middleware when building the middleware instead of checking that global variable in every request:
```
function rateLimitFn(userLimitsApi, endpointGroup = null) {
    if (!global.environment.enabledFeatures.rateLimitsEnabled) {
        return function noopRateLimitMiddleware (req, res, next) { next };
    }

    return function rateLimitMiddleware(req, res, next) {
        [...]
```

lib/cartodb/middleware/rate-limit.js
```diff
@@ -0,0 +1,69 @@
...
+    NAMED_LIST: 'named_list',
+    NAMED_CREATE: 'named_create',
+    NAMED_GET: 'named_get',
+    NAMED: 'named',
+    NAMED_UPDATE: 'named_update',
+    NAMED_DELETE: 'named_delete',
+    NAMED_TILES: 'named_tiles'
+};
+
+function rateLimitFn(userLimitsApi, endpointGroup = null) {
```
dgaubert: What about replacing `rateLimitFn` -> `rateLimit` and `rateLimit` -> `userRateLimit`??

lib/cartodb/api/user_limits_api.js
```diff
@@ -77,3 +87,145 @@ UserLimitsApi.prototype.getTimeoutRenderLimit = function (username, apiKey, call
...
+                self.rateLimits.redisCommand = 'EVAL';
+                return self.getRateLimit(user, endpointGroup, callback);
+            }
+
+            let rateLimit;
+            if (!err) {
+                rateLimit = getLowerRateLimit(rateLimits);
+            }
+            
+            callback(err, rateLimit);
```
dgaubert: Follow early error return pattern in order to be explicit as possible: 
```
if (err) {
  return callback(err);
}

const rateLimit = getLowerRateLimit(rateLimits);

callback(null, rateLimit);
```

lib/cartodb/api/user_limits_api.js
```diff
@@ -77,3 +87,145 @@ UserLimitsApi.prototype.getTimeoutRenderLimit = function (username, apiKey, call
...
+    `;
+}
+
+/**
+ * Returns the inner rateLimit what is the strictest one
+ * @param {Array} rateLimits Each inner array has 5 integers indicating: 
+ *      isBloqued, limit, remaining, retry, reset
+ */
+function getLowerRateLimit(rateLimits) {
+    /*jshint maxcomplexity:10 */
```
dgaubert: 😱

lib/cartodb/api/user_limits_api.js
```diff
@@ -77,3 +87,145 @@ UserLimitsApi.prototype.getTimeoutRenderLimit = function (username, apiKey, call
...
+    if (!rateLimits || !Array.isArray(rateLimits) || !rateLimits.length) {
```
dgaubert: it could be much simpler, couldn't it? `if (!Array.isArray(rateLimits)) { ... }`

lib/cartodb/api/user_limits_api.js
```diff
@@ -77,3 +87,145 @@ UserLimitsApi.prototype.getTimeoutRenderLimit = function (username, apiKey, call
...
+        return;
+    }
+
+    let minIndex = 0;
+    let minRemainingValue;
+    let currentIndex = 0;
+    for (let rateLimit of rateLimits) {
```
dgaubert: If you have to worry about incrementing `currentIndex`... why not use a `for` _de toda la vida_ ;)

lib/cartodb/api/user_limits_api.js
```diff
@@ -77,3 +87,145 @@ UserLimitsApi.prototype.getTimeoutRenderLimit = function (username, apiKey, call
...
+
+        if (minRemainingValue === undefined || remaining < minRemainingValue) {
+            minIndex = currentIndex;
+            minRemainingValue = remaining;
+        }
+
+        currentIndex++;
+    }
+
+    if (validateRatelimit(rateLimits[minIndex])) {
```
dgaubert: Didn't you validate it on line 199??

lib/cartodb/api/user_limits_api.js
```diff
@@ -77,3 +87,145 @@ UserLimitsApi.prototype.getTimeoutRenderLimit = function (username, apiKey, call
...
+    var self = this;
+
+    let redisParams = [
+        this.rateLimits.redisCommand === 'EVAL' ? this.rateLimits.lua : this.rateLimits.sha,
+        2,
+        getStoreKey(user, endpointGroup),   // KEY[1] 
+        getStatusKey(user, endpointGroup)   // KEY[2]
+    ];
+
+    this.metadataBackend.redisCmd(
```
dgaubert: Why not move this function to `cartodb-redis` and manage all the Redis stuff there??

Then you can use:

```
this.metadataBackend.getRateLimit(user, endpointGroup, function (err, rateLimits) { ... });
```

lib/cartodb/api/user_limits_api.js
oleurud: It is not the same. I can remove the first because `Array.isArray` returns false to `null` or `undefined` values, but not the last, because I want to know `rateLimits.length` is bigger than `0`

``` !Array.isArray(rateLimits) || !rateLimits.length ```

File: lib/cartodb/api/user_limits_api.js (hunk 21 above)
Comment (oleurud) [reply lib/cartodb/api/user_limits_api.js]: Sounds well ;)

File: lib/cartodb/api/user_limits_api.js (hunk 17 above)
Comment (oleurud) [reply lib/cartodb/api/user_limits_api.js]: I can refactor with a new function but I don't want it.

File: lib/cartodb/middleware/rate-limit.js [hunk 22]
```diff
@@ -0,0 +1,64 @@
...
+            }
+
+            return next();
+        });
+    };
+}
+
+
+function isRateLimitEnabled(endpointGroup) {
+    return global.environment.enabledFeatures.rateLimitsEnabled &&
```
enekid: Every time I see this `global` variable, a snail dies.

lib/cartodb/controllers/named_maps_admin.js
```diff
@@ -26,6 +29,7 @@ NamedMapsAdminController.prototype.register = function (app) {
         userMiddleware(),
         localsMiddleware(),
         credentialsMiddleware(),
+        rateLimit(this.userLimitsApi, RATE_LIMIT_ENDPOINTS_GROUPS.NAMED_CREATE),
```
dgaubert: Is there any reason to place `rateLimit` after `credentials`? Shall we put it just after `userMiddleware`?

lib/cartodb/middleware/rate-limit.js
oleurud: Me too

lib/cartodb/controllers/named_maps_admin.js
oleurud: We have talked about `authorization` or `rate limit` first, and finally, we will continue with auth first

dgaubert: Actually, other controllers have `rateLimit` first and then `auth` and this controller you placed `rateLimit` between `credentials` and `authorizedByAPIKey `.

@enekid and I were talking about the order and we realized that having the fallback system all requests are automatically authorized (`publicuser` by default). So until we don't remove the fallback to old auth system the right order should be: `rateLimit` first and then `auth`.

oleurud: Middleware order fixed

PR #877, opened 2018-02-19
Auth API

dgaubert:


lib/cartodb/backends/pg_connection.js
```diff
@@ -20,45 +18,90 @@ module.exports = PgConnection;
...
-            user_params.user_id = user_id;
-            var dbuser = _.template(auth_user, user_params);
-            _.extend(params, {dbuser:dbuser});
-
-            // skip looking up user_password if postgres_auth_pass
-            // doesn't contain the "user_password" label
-            if (!auth_pass || ! auth_pass.match(/\buser_password\b/) ) {
-                return null;
+PgConnection.prototype.setDBAuth = function(username, params, apikeyType, callback) {
+    if (apikeyType === 'master') {
```
dgaubert: As we said before: _this is ugly as hell but first, let's have it working and then we can apply some best practices_. We have the following pattern:
```
if (type === 'foo') {
  ...
} else if (type === 'bar') {
  ...
} else if (type === '...') {
  ...
} else {
  ...
}
```
We should apply [Replace Conditional with Polymorphism](https://refactoring.com/catalog/replaceConditionalWithPolymorphism.html). The steps to refactor this piece of code  could be the following (with better names and so on):
 - Create a class hierarchy to implement the same interface, for each type, create a subclass:
   + `Credentials.getDBAuth()` <- `RegularCredentials.getDBAuth()`
   + `Credentials.getDBAuth()` <- `MasterCredentials.getDBAuth()`
   + `Credentials.getDBAuth()` <- `DefaultCredentials.getDBAuth()`
 - Create `CredentialsType` that is in charge to use the right instance of `Credentials` and set the right credential for the `api-key`

HTH.

dgaubert: We did something similar [here](https://github.com/CartoDB/Windshaft-cartodb/tree/master/lib/cartodb/backends/layer-stats). Although, I think we went so far.

enekid: We will tackle this when we remove the auth fallback

lib/cartodb/middleware/context/apikey-credentials.js
```diff
@@ -0,0 +1,84 @@
+'use strict';
+
+module.exports = () => function apikeyTokenMiddleware(req, res, next) {
```
dgaubert: @enekid, Following your suggestions in previous CR: this arrow function binds the `module` context to this middleware. I'd prefer to have its own context unaltered.

lib/cartodb/api/auth_api.js
```diff
@@ -47,39 +46,113 @@ AuthApi.prototype.authorizedBySigner = function(res, callback) {
...
-    var givenKey = req.query.api_key || req.query.map_key;
-    if ( ! givenKey && req.body ) {
-        // check also in request body
-        givenKey = req.body.api_key || req.body.map_key;
-    }
-    if ( ! givenKey ) {
-        return callback(null, 0); // no api key, no authorization...
+AuthApi.prototype.authorizedByAPIKey = function(user, res, callback) {
+    const apikeyToken = res.locals.api_key;
+    const apikeyUsername = res.locals.apikeyUsername;
```
dgaubert: Why not call `const apikeyCredentials = getApikeyCredentialsFromRequest(req);` directly from here and use `token` as `api_key` and `username` as `user`?. So we don't need to call it from a separated middleware (`apikeyTokenMiddleware`) since we are using it only here. Then we can remove that middleware.

lib/cartodb/middleware/context/apikey-credentials.js
dgaubert: This middleware is not doing so much. In fact `apikeyUsername` is only used in `auth_api`.  So I'd remove this module as middleware and export `getApikeyCredentialsFromRequest ` function since it's the actual function that we need to use.

enekid: OK

lib/cartodb/api/auth_api.js
enekid: It's made this way by design. The final order of rate-limit/authorization is not yet settled or could change in the future. This way we separate the 2 concerns: 
- extracting credentials/api key token
- proper authorization

We could insert the rate-limit middleware between those middlewares in case we would like to rate-limit based on api key.

lib/cartodb/backends/pg_connection.js
```diff
@@ -109,28 +141,37 @@ PgConnection.prototype.setDBConn = function(dbowner, params, callback) {
  */
 
 PgConnection.prototype.getConnection = function(username, callback) {
-    var self = this;
+    require('debug')('cachechan')("getConn1");
```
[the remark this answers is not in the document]
enekid: OK!

lib/cartodb/api/auth_api.js
dgaubert: > it's made this way by design. The final order of rate-limit/authorization is not yet settled or could change in the future.

In general, foreseeing needs leads to over-engineering. If anyone has to implement a different middleware to save credentials in `res.locals` to be used afterward, is the right one to do it.

By the way, I've seen that the change is not easy because you need `request` object in `authorizedByAPIKey()`. So, fair enough let's move on.

lib/cartodb/backends/pg_connection.js
```diff
@@ -71,36 +110,30 @@ PgConnection.prototype.setDBAuth = function(username, params, callback) {
 // @param callback function(err)
 //
 PgConnection.prototype.setDBConn = function(dbowner, params, callback) {
-    var self = this;
-    // Add default database connection parameters
-    // if none given
     _.defaults(params, {
-        dbuser: global.environment.postgres.user,
-        dbpassword: global.environment.postgres.password,
+        // dbuser: global.environment.postgres.user,
```
dgaubert: Do not comment code. Better remove it.

enekid: Yep. I missed that.

PR #880, opened 2018-02-23
Postgis permission error

enekid:
Reformat `Postgis permission denied for relation` error, removing technical information.

- set error message as: `'permission denied for relation'`
- type as `'auth'` if none set before (eg. `'layer'`)
- add info field with offending relation name

### DIFF
--- a/lib/cartodb/middleware/error-middleware.js
+++ b/lib/cartodb/middleware/error-middleware.js
@@ -62,9 +62,9 @@ const offendingRelationRegex = /permission denied for relation (\w+)/i;
 
 function getOffendingRelationFromError(err) {
     if (err.message) {
-        const relation = offendingRelationRegex.exec(err.message);
-        if (relation && relation.length > 1) {
-            return relation[1];
+        const relations = offendingRelationRegex.exec(err.message);
+        if (relations && relations.length > 1) {
+            return relations[1];
         }
     }
     return null;
--- a/lib/cartodb/middleware/error-middleware.js
+++ b/lib/cartodb/middleware/error-middleware.js
@@ -88,7 +88,6 @@ function populateErrors (errors) {
 
         if (isPermissionDeniedForRelationError(error)) {
             error.info = getOffendingRelationFromError(error);
-            error.message = 'permission denied for relation';
             error.type = error.type || 'auth';
             error.subtype = 'permission-denied-relation';
             error.http_status = 403;
--- a/lib/cartodb/middleware/error-middleware.js
+++ b/lib/cartodb/middleware/error-middleware.js
@@ -54,6 +54,22 @@ function isTimeoutError (err) {
     return isRenderTimeoutError(err) || isDatasourceTimeoutError(err);
 }
 
+function isPermissionDeniedForRelationError(err) {
+    return err.message && err.message.match(/permission denied for relation/i);   
+}
+
+const offendingRelationRegex = /permission denied for relation (\w+)/i;
+
+function getOffendingRelationFromError(err) {
+    if (err.message) {
+        const relation = offendingRelationRegex.exec(err.message);
+        if (relation && relation.length > 1) {
+            return relation[1];
+        }
+    }
+    return null;
+}
+
 function populateTimeoutErrors (errors) {
     return errors.map(function (error) {
         if (isRenderTimeoutError(error)) {
@@ -68,6 +84,17 @@ function populateTimeoutErrors (errors) {
             error.message = 'You are over platform\'s limits. Please contact us to know more details';
             error.type = 'limit';
             error.http_status = 429;
+        }
+
+        if (isPermissionDeniedForRelationError(error)) {
+            error.info = getOffendingRelationFromError(error);
+            error.message = 'permission denied for relation';
+            error.type = error.type || 'auth';
+            error.subtype = 'permission-denied-relation';
+            error.http_status = 403;
+
+            // console.log('error');
+            // console.log(error);
         }
 
         return error;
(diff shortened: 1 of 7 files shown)


lib/cartodb/middleware/error-middleware.js
```diff
@@ -70,6 +86,14 @@ function populateTimeoutErrors (errors) {
             error.http_status = 429;
         }
 
+        if (isPermissionDeniedForRelationError(error)) {
+            error.info = getOffendingRelationFromError(error);
```
dgaubert: Is that change agreed with frontend people?

lib/cartodb/middleware/error-middleware.js
```diff
@@ -54,6 +54,22 @@ function isTimeoutError (err) {
...
+    if (err.message) {
+        const relation = offendingRelationRegex.exec(err.message);
+        if (relation && relation.length > 1) {
+            return relation[1];
+        }
+    }
+    return null;
+}
+
 function populateTimeoutErrors (errors) {
```
dgaubert: We should change the name of this function, now populated two type of errors: timeout and auth.

lib/cartodb/middleware/error-middleware.js
```diff
@@ -54,6 +54,22 @@ function isTimeoutError (err) {
...
 
+function isPermissionDeniedForRelationError(err) {
+    return err.message && err.message.match(/permission denied for relation/i);   
+}
+
+const offendingRelationRegex = /permission denied for relation (\w+)/i;
+
+function getOffendingRelationFromError(err) {
+    if (err.message) {
+        const relation = offendingRelationRegex.exec(err.message);
```
dgaubert: being a PITA 👿 : it's actually an array of matches so the name should be plural

test/support/test-client.js
```diff
@@ -47,6 +47,12 @@ module.exports.RESPONSE = {
         headers: {
             'Content-Type': 'application/json; charset=utf-8'
         }
+    },
+    GENERAL_AUTH_ERROR: {
```
dgaubert: `AUTH_ERROR` looks better. doesn't it?

lib/cartodb/middleware/error-middleware.js
enekid: Postgres gives only the name of the first relation that rises the error.

For example, with this query, and having no access granted to neither `table1` nor `table2`:

`select *, (SELECT count(1) from table1) as tc1 from table2`

Postgis would return this error:

`Error: Postgis Plugin: ERROR:  permission denied for relation table2 in executeQuery Full sql was: ...`

dgaubert: Yeah, I know. I'm talking about `regex` API. The `.exec(`method returns an array of matches. And then we access to the very first element of that array `relation[1]` which the thing that sounds a little weird. As a convention, we use plurals to give a name to `arrays`. For instance: `relations[1]` or `const relation = mathes[1]`.

test/support/test-client.js
enekid: There're are 2 types of auth errors:
1. `401` API key not found
2. `403` All other auth errors:
    - _api_ access not granted
    - _dataset_ access not granted

I've created 2 types of auth errors:
- `GENERIC_AUTH_ERROR` -> 403
- `APIKEY_NOT_FOUND_AUTH_ERROR` -> 401

test/support/test-client.js
```diff
@@ -47,6 +47,18 @@ module.exports.RESPONSE = {
         headers: {
             'Content-Type': 'application/json; charset=utf-8'
         }
+    },
+    GENERIC_AUTH_ERROR: {
```
dgaubert: why not use `UNAUTHORIZED`?

test/support/test-client.js
```diff
@@ -47,6 +47,18 @@ module.exports.RESPONSE = {
...
+        status: 403,
+        headers: {
+            'Content-Type': 'application/json; charset=utf-8'
+        }
+    },
+    APIKEY_NOT_FOUND_AUTH_ERROR: {
```
dgaubert: `FORBIDDEN` ??

test/support/test-client.js
```diff
@@ -47,6 +47,18 @@ module.exports.RESPONSE = {
         headers: {
             'Content-Type': 'application/json; charset=utf-8'
         }
+    },
+    UNAUTHORIZED_ERROR: {
```
dgaubert: 403 -> `FORBIDDEN_ERROR `

test/support/test-client.js
```diff
@@ -47,6 +47,18 @@ module.exports.RESPONSE = {
...
+        status: 403,
+        headers: {
+            'Content-Type': 'application/json; charset=utf-8'
+        }
+    },
+    FORBIDDEN_ERROR: {
```
dgaubert: 401 -> `UNAUTHORIZED_ERROR`

test/support/test-client.js
enekid: ¯\_(ツ)_/¯ Thanks!

test/support/test-client.js
enekid: ¯\_(ツ)_/¯ Thanks!

lib/cartodb/middleware/error-middleware.js
dgaubert: if not, I wouldn't add it.

dgaubert:
Shall we merge this PR?

enekid:
I would put this on hold until we revamp the response error system.

dgaubert:
What does it mean `revamp the response error system`? I'm missing something ¿?

enekid:
This PR doesn't make much sense on it's own until we refactor the whole error management system.

PR #882, opened 2018-03-02
Middleware refactor

dgaubert:
Some minor refactor in order to follow same patterns.


lib/cartodb/middleware/context/layergroup-token.js
```diff
@@ -1,32 +1,33 @@
 var LayergroupToken = require('../../models/layergroup-token');
 
-module.exports = function layergroupTokenMiddleware(req, res, next) {
-    if (!res.locals.token) {
-        return next();
-    }
+module.exports = function layergroupToken () {
+    return function layergroupTokenMiddleware(req, res, next) {
```
enekid: Missing space before parameter list 🙀 
```javascript
return function layergroupTokenMiddleware (req, res, next) {`
```

lib/cartodb/middleware/context/layergroup-token.js
```diff
@@ -1,32 +1,33 @@
...
-        } else if (res.locals.signer !== user) {
-            var err = new Error(`Cannot use map signature of user "${res.locals.signer}" on db of user "${user}"`);
-            err.type = 'auth';
-            err.http_status = 403;
-            if (req.query && req.query.callback) {
-                err.http_status = 200;
+        if (layergroupToken.signer) {
+            res.locals.signer = layergroupToken.signer;
+            if (!res.locals.signer) {
+                res.locals.signer = user;
```
enekid: Unreachable code: `res.locals.signer = user;`
The *if* clause `if (!res.locals.signer) {` has to be removed.
`res.locals.signer` has some _truthy_ value since  it's equal to `layergroupToken.signer` and inside this block `layergroupToken.signer` *must* have a _truthy_ value.

dgaubert: 👍  In fact, we fixed that in another branch if I recall correctly.

PR #883, opened 2018-03-02
Do not force format for default named tiles

dgaubert:
Fixes https://github.com/CartoDB/support/issues/1351

### DIFF
--- a/lib/cartodb/controllers/named_maps.js
+++ b/lib/cartodb/controllers/named_maps.js
@@ -89,7 +89,7 @@ NamedMapsController.prototype.register = function(app) {
     );
 };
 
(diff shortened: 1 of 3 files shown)


lib/cartodb/controllers/named_maps.js
```diff
@@ -80,16 +89,16 @@ NamedMapsController.prototype.register = function(app) {
     );
 };
 
-NamedMapsController.prototype.getNamedMapProvider = function (label) {
+NamedMapsController.prototype.getNamedMapProvider = function ({ label, forcedFormat }) {
```
enekid: Indicate that `forceFormat` param is optional:

```javascript
NamedMapsController.prototype.getNamedMapProvider = function ({ label, forcedFormat = null }) {
```

Issue #497, opened 2016-06-07
Null category with 0 count


rochoa:
![687474703a2f2f692e696d6775722e636f6d2f43396473504b622e706e67](https://cloud.githubusercontent.com/assets/45346/14313819/1dcc8522-fbf4-11e5-968f-61d662712e42.png)

![687474703a2f2f692e696d6775722e636f6d2f666142527578322e706e67](https://cloud.githubusercontent.com/assets/45346/14313834/2d662ed4-fbf4-11e5-879d-f578b23d236f.png)

ref https://github.com/CartoDB/deep-insights.js/issues/269

replaces https://github.com/CartoDB/Windshaft/issues/468

jorgesancha:
Adding to TODO column in Engine

dgaubert:
I've just pushed a [test](https://github.com/CartoDB/Windshaft-cartodb/commit/f5f0601e53cd5cd90f46713fba12595396e10446) to check if category NULL count its values properly. It just passed. :man_shrugging:

@AbelVM, could you share your dataset to take a look?

AbelVM:
6 months later... Nope, I'm afraid I don't have the dataset above :disappointed:

dgaubert:
No problem, thanks for your patient...

dgaubert:
Ready for CR in #617

ethervoid:
@dgaubert could you please add some acceptance steps to pass it? thank you!

dgaubert:
Sure thing!

1. Go to cartodb-staging.com
2. Create a map using populated_places dataset.
3. Create a category widget with adm0name as aggregation column and count operation:
![screen shot 2017-01-19 at 16 38 41](https://cloud.githubusercontent.com/assets/1936888/22113207/d1cd580e-de65-11e6-9f50-15dca03ba9eb.png)
Should look like:
![screen shot 2017-01-19 at 16 40 50](https://cloud.githubusercontent.com/assets/1936888/22113308/1bad0cc6-de66-11e6-8957-80391eaf2632.png)
4. Change operation to SUM over `pop_max`column, should look like:
![screen shot 2017-01-19 at 16 57 43](https://cloud.githubusercontent.com/assets/1936888/22114063/8581f40c-de68-11e6-810a-ac454bba6ef9.png)
[...]

PR #888, opened 2018-03-08
Aggregation count: Do not return null categories

Algunenano:
Related to https://github.com/CartoDB/support/issues/1368

The issue appeared only if the `aggregationColumn` wasn't provided, and as it's optional in `count ` aggregation, it is being provided in the main map screen but not in the embeds.

### DIFF
--- a/lib/cartodb/models/dataview/aggregation.js
+++ b/lib/cartodb/models/dataview/aggregation.js
@@ -42,7 +42,7 @@ const rankedCategoriesQueryTpl = ctx => `
             ${ctx.aggregationFn} AS value,
             row_number() OVER (ORDER BY ${ctx.aggregationFn} desc) as rank
         FROM (${filteredQueryTpl(ctx)}) filtered_source
-        ${ctx.aggregationColumn !== null ? `WHERE ${ctx.aggregationColumn} IS NOT NULL` : ''}
+        WHERE ${ctx.aggregation === "count" ? `${ctx.column}` : `${ctx.aggregationColumn}`} IS NOT NULL
         GROUP BY ${ctx.column}
         ORDER BY 2 DESC
     )
@@ -279,7 +279,7 @@ module.exports = class Aggregation extends BaseDataview {
             max_val = 0,
             categories_count = 0
          } = result.rows[0] || {};
-        
+
         return {
             aggregation: this.aggregation,
             count: count,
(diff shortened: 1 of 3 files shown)


Algunenano:
Ok, seeing https://github.com/CartoDB/Windshaft-cartodb/issues/497 seems that returning a NULL category with 'count' was done in purpose. Needs to be decided then

Algunenano:
The decision is to remove the `NULL` categories in all cases, so I've updated the test that checked it and added some extra test to check that both cases (with and without `aggregationColumn`) don't return NULL values.

test/acceptance/dataviews/aggregation.js
```diff
@@ -425,3 +416,79 @@ describe('aggregation dataview tuned by categories query param', function () {
...
+        this.testClient = new TestClient(mapConfig, 1234);
+        this.testClient.getDataview('categories', { own_filter: 0, categories: 0 }, (err, dataview) => {
+            assert.ifError(err);
+            assert.equal(dataview.categories.length, 3);
+            this.testClient.drain(done);
+        });
+    });
+
+    it(`should handle null values correctly when aggregationColumn is provided`, function (done) {
+        mapConfig.dataviews.categories.options.aggregationColumn = 'val';
```
oleurud: Only a detail. I would prefer to add the aggregation params as a parameter of a function, making future tests independent.

```
function mapConfig (aggregationOptions = {column: 'cat', aggregation: 'count'})
```

lib/cartodb/models/dataview/aggregation.js
```diff
@@ -42,7 +42,7 @@ const rankedCategoriesQueryTpl = ctx => `
             ${ctx.aggregationFn} AS value,
             row_number() OVER (ORDER BY ${ctx.aggregationFn} desc) as rank
         FROM (${filteredQueryTpl(ctx)}) filtered_source
-        ${ctx.aggregationColumn !== null ? `WHERE ${ctx.aggregationColumn} IS NOT NULL` : ''}
+        WHERE ${ctx.aggregation === "count" ? `${ctx.column}` : `${ctx.aggregationColumn}`} IS NOT NULL
```
oleurud: :+1:

test/acceptance/dataviews/aggregation.js
Algunenano: It's a nice idea, but the only aggregation where this matters is `count` since the rest are invalid if they don't provide an `aggregationColumn`. The others are also tested above.

TBH this test repeats a test above an only the one without `aggregationColumn` is needed, but since the behaviour differed I prefer to have both here together.

PR #891, opened 2018-03-09
Refactor map controler

dgaubert:
Applies some changes in order to separate concerns and make easier further refactor


dgaubert:
🙏 @oleurud & @enekid

PR #899, opened 2018-03-13
Named Maps Admin: Extract middlewares form controller's context

dgaubert:


lib/cartodb/controllers/named_maps_admin.js
```diff
@@ -167,49 +172,56 @@ NamedMapsAdminController.prototype.retrieve =  function () {
...
-            const method = req.query.callback ? 'jsonp' : 'json';
-            res[method]({ template_ids: templateIds });
+            next();
         });
-    }.bind(this);
-};
+    };
+}
+
+function sendResponse () {
```
oleurud: :heart_decoration:

lib/cartodb/controllers/named_maps_admin.js
```diff
@@ -78,10 +83,23 @@ NamedMapsAdminController.prototype.register = function (app) {
     );
 };
 
-NamedMapsAdminController.prototype.authorizedByAPIKey = function (action, label) {
+function checkContentType (action, label) {
```
enekid: Wouldn't it to be better to send objects instead of a list of parameters?
```javascript
function checkContentType ({action, label}) {}
```
I've seen it in other controllers and I like it.

dgaubert: sure, and for the sake of consistency...

PR #901, opened 2018-03-14
Refactor named maps controller

dgaubert:
In order to make middlewares easily interchangeable, I extracted them from controller's context.

### DIFF
--- a/lib/cartodb/controllers/named_maps.js
+++ b/lib/cartodb/controllers/named_maps.js
@@ -80,7 +80,7 @@ NamedMapsController.prototype.register = function(app) {
         getAffectedTables(),
         getTemplate(staticOptions),
         prepareLayerFilterFromPreviewLayers(staticOptions),
-        this.getStaticImageOptions(),
+        getStaticImageOptions({ tablesExtentApi: this.tablesExtentApi }),
         this.getImage(staticOptions),
         this.incrementMapViews(),
         this.setSurrogateKey(),
@@ -116,7 +116,7 @@ function getNamedMapProvider ({ namedMapProviderCache, label, forcedFormat = nul
             next();
         });
     };
-};
+}
 
 function getAffectedTables () {
     return function getAffectedTables (req, res, next) {
@@ -134,7 +134,7 @@ function getAffectedTables () {
             next();
         });
     };
-};
+}
 
 function getTemplate ({ label }) {
     return function getTemplateMiddleware (req, res, next) {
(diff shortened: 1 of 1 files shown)


lib/cartodb/controllers/named_maps.js
```diff
@@ -50,19 +52,21 @@ NamedMapsController.prototype.register = function(app) {
...
-        this.getTile(tileOptions),
-        this.setSurrogateKey(),
-        this.setCacheChannelHeader(),
-        this.setLastModifiedHeader(),
-        this.setCacheControlHeader(),
-        this.setContentTypeHeader(),
-        this.respond(),
+        getNamedMapProvider(tileOptions),
+        getAffectedTables(),
+        getTile(tileOptions),
```
oleurud: maybe it is more clear to send the parameters needed:
```
getTile(this.tileBackend, 'NAMED_MAP_TILE')
```

lib/cartodb/controllers/named_maps.js
```diff
@@ -73,23 +77,23 @@ NamedMapsController.prototype.register = function(app) {
...
-        this.setCacheChannelHeader(),
-        this.setLastModifiedHeader(),
-        this.setCacheControlHeader(),
-        this.setContentTypeHeader(),
-        this.respond()
+        getNamedMapProvider(staticOptions),
+        getAffectedTables(),
+        getTemplate(staticOptions),
+        prepareLayerFilterFromPreviewLayers(staticOptions),
+        getStaticImageOptions({ tablesExtentApi: this.tablesExtentApi }),
```
oleurud: I am losing something, but why send objects and then destructuring?

dgaubert: There are a couple of reasons:
  - make optional parameters much less PITA.
  - unorder params identified by name.
  - avoid long params list
  -  you can still define default values
  - ...

lib/cartodb/controllers/named_maps.js
dgaubert: sure

lib/cartodb/controllers/named_maps.js
oleurud: Agree, but even with only one?

oleurud: And it does not seem bad to me, it just makes me a little weird

enekid: Look at this as a way to pass `named` parameters.

dgaubert: > Agree, but even with only one?

To keep consistency and to be ready to add new params

PR #907, opened 2018-03-15
Update camshaft to 0.61.5

Algunenano:
Related to https://github.com/CartoDB/support/issues/1389 and https://github.com/CartoDB/support/issues/1392


Algunenano:
Checked that it works correctly for AOI but I'll create a PR which fixes JOIN too

PR #909, opened 2018-03-19
Spread _prepareContext_ middleware

dgaubert:
This PR includes:
- Removes _meta-middleware_ `prepareContext`.
- Reorder involved middlewares and use them explicitly for each controller
- Merge `allowQueryParams` and `cleanUpQueryParams` middlewares
- Fix ported tests by adding `host` header to indicate the user in the `req/res` cycle and remove the old `req2params` middleware.


lib/cartodb/controllers/named_maps.js
```diff
@@ -222,9 +238,11 @@ function getTile ({ tileBackend, label }) {
                 return next(err);
             }
 
-            res.locals.body = tile;
-            res.locals.headers = headers;
-            res.locals.stats = stats;
```
oleurud: don't we need to save the stats?

dgaubert: We were doing `req.profiler.add(stats);` twice. Here and in `sendResponse` middleware.

oleurud: Oki!

lib/cartodb/controllers/named_maps.js
```diff
@@ -222,9 +238,11 @@ function getTile ({ tileBackend, label }) {
...
+            if (headers) {
+                res.set(headers);
+            }
+
+            res.body = tile;
```
oleurud: :heart:

PR #913, opened 2018-03-21
[DO NOT MERGE] Aggregation filters

jgoizueta:
This is to support the cartogl library

### DIFF
--- a/docs/aggregation.md
+++ b/docs/aggregation.md
@@ -185,3 +185,80 @@ This is the minimum number of (estimated) rows in the dataset (query results) fo
     ]
 }
 ```
+
+### `filters`
+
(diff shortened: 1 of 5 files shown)


jgoizueta:
Hey @dgaubert  @enekid  this is finally ready from some CR!

Note about the last 5 commits:
* First one fixes the problem with aggregation columns in ST_AsMVT and next-to-last one add tests for it (yeah, I now, bad order).
* The second and third fix a couple of unrelated problems revealed while preparint the MVT tests (I'ts always good to add tests 😅 )
* And the sweet final one fixes another problem with the MVT aggregation columns revealed by the test, yay!!

PR #917, opened 2018-03-23
Locals middleware refactor

dgaubert:
- All controller's collaborators no longer receive `res.locals`. Instead, they receive an object with only params they actually need.
- Extract `dbParamsFromReqParams`method to a module.
- Handle `Tile does not exist` for `mvt` format in `tileError` middleware instead of doing that in the generic error middleware.


lib/cartodb/utils/database-params.js
```diff
@@ -0,0 +1,25 @@
+module.exports = function getDatabaseConnectionParams (params) {
```
oleurud: :heart:

PR #918, opened 2018-03-23
Layergroup split

dgaubert:
- Reorder `rateLimit` and `cleanUpQeuryParams` Middleware
- Split layergroup controllers into small controllers: `tile`, `attributes`, `static`, `dataviews` and `analysis status`.


lib/cartodb/controllers/layergroup/analysis.js
```diff
@@ -0,0 +1,75 @@
...
+    constructor (
+        analysisStatusBackend,
+        pgConnection,
+        mapStore,
+        userLimitsApi,
+        layergroupAffectedTablesCache,
+        authApi,
+        surrogateKeysCache
+    ) {
+        this.analysisStatusBackend = analysisStatusBackend;
```
enekid: Would it make sense to use `Object.assign` in these constructors? Probably not, but I don't like so much repetition.

Something like:

```javascript
Object.assign(this, 
  pgConnection,
  mapStore,
  userLimitsApi,
  layergroupAffectedTablesCache,
  authApi,
  surrogateKeysCache,
);
```

lib/cartodb/controllers/layergroup/dataview.js
```diff
@@ -0,0 +1,199 @@
+const cors = require('../../middleware/cors');
```
enekid: Shouldn't we add `'use strict;'` in all new files?

lib/cartodb/controllers/layergroup/static.js
```diff
@@ -0,0 +1,161 @@
...
+            surrogateKeyHeader({ surrogateKeysCache: this.surrogateKeysCache }),
+            lastModifiedHeader(),
+            sendResponse()
+        );
+    }
+};
+
+function getPreviewImageByCenter (previewBackend) {
+    return function getPreviewImageByCenterMiddleware (req, res, next) {
+        const width = +req.params.width;
```
enekid: Why do we put these `+` operators? I have forgotten :(

lib/cartodb/controllers/layergroup/dataview.js
dgaubert: Since I started to write ES6 code with modules I get rid it of my mind. So we should still put `use strict` directive, a PR fixing it will be more than welcome.

lib/cartodb/controllers/layergroup/analysis.js
dgaubert: I have mixed feeling about this. I prefer being explicit in some cases and I' m not totally sure is totally safe using `Object.assing` to link object dependencies. If you have some literature about this topic it would be much appreciated.

lib/cartodb/controllers/layergroup/static.js
dgaubert:
```bash
$ node
> let width = '200'
undefined
> width
'200'
> width = +width
200
> width
200
>
```

enekid: Now I remember!
This cheatsheet is great for dealing with number castings:
https://i.stack.imgur.com/dzin3.png

Thanks!

lib/cartodb/controllers/layergroup/analysis.js
enekid: Ok!

PR #920, opened 2018-03-26
Error with aggregations and MVT Postgis backend

rochoa:


rochoa:
@jgoizueta, after updating this branch with `master`, the test is passing. I still think it's worthy to incorporate this test to the suite. What do you think?

test/acceptance/mvt-regressions.js
```diff
@@ -0,0 +1,78 @@
...
+
+            after(function (){
+                serverOptions.renderer.mvt.usePostGIS = originalUsePostGIS;
+            });
+
+            afterEach(function (done) {
+                this.testClient.drain(done);
+            });
+
+            it('aggregates with centroid placement', function (done) {
```
dgaubert: In description, we want a centroid placement but the mapconfig specifies `point-sample`.

jgoizueta:
I think your test had to be failing because of this: https://github.com/CartoDB/Windshaft-cartodb/pull/913/commits/2132960d7c372c3d2bbf0820dd85dc7ceb8b5c98

(`pop_max` was being added to `columns` instead of `pop_max_avg`)

I added tests for that here: https://github.com/CartoDB/Windshaft-cartodb/pull/913/commits/3d36802686aa9d618a92dade88f939b13fa2e01f

So that should be covered there but I'd add your test too (I see I didn't add tests without dimensions)

PR #926, opened 2018-04-03
Do not log analysis results to stdout when log filename is not provided

dgaubert:
To remove the noise that analysis logger produces when running test: do not log analysis results to stdout when log filename is not provided

### DIFF
--- a/lib/cartodb/backends/analysis.js
+++ b/lib/cartodb/backends/analysis.js
@@ -48,9 +48,11 @@ AnalysisBackend.prototype.create = function(analysisConfiguration, analysisDefin
     analysisConfiguration.batch.inlineExecution = this.batchConfig.inlineExecution;
     analysisConfiguration.batch.hostHeaderTemplate = this.batchConfig.hostHeaderTemplate;
 
-    analysisConfiguration.logger = {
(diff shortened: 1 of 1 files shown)


Algunenano:
Does this affect running the app too or just the test? I run the service with `filename: undefined` so the logs output to stdout are automatically handled by systemd/journalctl. I guess this might break that.

Also, from the config file:
From the config file:
```
            // If filename is given logs comming from analysis client  will be written
            // there, in append mode. Otherwise 'log_filename' is used. Otherwise stdout is used (default).
            // Log file will be re-opened on receiving the HUP signal
```

Algunenano:
> What do you do for application's logs?
Everything is output to `stdout` but since it's being run through a systemd process the output is captured and stored with `journalctl`so I can do things `journalctl --since=yesterday -u carto-tiler.service`.

Could we instead have `analysis.logger` set to something like `test-analysis.log` in `environments/test.js.example` and add that file to the .gitignore? Would that solve your issue?

dgaubert:
Closing this PR. 

Thanks for the help. I was the one who did analyses logging and I didn't see the way to go. :(

PR #932, opened 2018-04-06
Fix layergroupid date

oleurud:
Fix: https://github.com/CartoDB/support/issues/1365

We are returning the same layergroupid date after a data update.

### DIFF
--- a/lib/cartodb/controllers/map.js
+++ b/lib/cartodb/controllers/map.js
@@ -422,7 +422,7 @@ function setLastUpdatedTimeToLayergroup () {
         const { mapConfigProvider, analysesResults } = res.locals;
         const layergroup = res.body;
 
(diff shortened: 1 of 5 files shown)


rochoa:
BTW, why don't we create a test for this regression?

dgaubert:
Yep, we have to add regression test here. We didn't create it on Friday because we wanted to deploy y the fix ASAP and the test is not trivial. The test would look like:
 1) Create an anonymous map.
 2) Check `cache-buster` in the layergroup response.
 3) Update (by adding new rows) of one of the table-
 4) Instantiate the same map again
 5) Check the `cache-buster` is updated.

@oleurud ping me if you need help here.

dgaubert:
Let me take care of it since @oleurud is working on another issue.

PR #934, opened 2018-04-09
Improve folder structure

dgaubert:
The target of this PR is to reflect in the folder structure the new application logic, done in #931.


lib/cartodb/api/map/dataview-layergroup-controller.js
```diff
@@ -1,15 +1,15 @@
-const layergroupToken = require('../../middleware/layergroup-token');
```
enekid: Just a proposal to these long/moving require paths. If we stick to the unique middlewares directory we could use one of this options (o even both of them):

# process.env.middlewaresDir

At bootstrapping, set a global constant with the middlewares directory path
```javascript
process.env.middlewaresDir = path.join(__dirname, 'wadus/middlewares')
```

Use this constant to require the middlewares:
```javascript
const credentials = require(process.env.middlewaresDir + '/credentials');
```

# Middlewares object

Create an `index.js` in the middlewares directory exporting an object with all the middlewares:
```javascript
module.export = {
  credentials = require('./credentials'),
  dbConnSetup = require('./db-conn-setup')
}
```
Then you can import this object and use the middlewares you need:

```javascript
const middlewares = requires('wadus/middlewares');
const credentials  = middlewares.credentials;
const dbConnSetup = middleares.dbConnSetup;
```

dgaubert: Let's how it evolves. 

BTW, For years Node.js developers have tried to solve long paths for `requires`. I've tried some solutions but no one seems to be the right way (symlinks, globals, basePaths, npm folders)

dgaubert: Nowadays this is my preferred solution: https://medium.com/@arnaudrinquin/build-modular-application-with-npm-local-modules-dfc5ff047bcc

Applied to this context, `package.json`: 

```
{
 “name”: “windshaft-cartodb”,
 “dependencies”: {
   “@middlewares”: “file:lib/cartodb/api/middlewares”,
 }
}
```
And then, in somewhere:
```
const { credentials } = require('@middlewares');
```

oleurud: I have used `The Container` and a similar `The Global` solutions. Both work fine, but I prefer `The Global` because for me it is more JS adapted.

I haven't seen before the `local package` feature, I love it.

dgaubert: It has the problem you need to add an extra pass to remove locals dependencies from `node_modules` when you modify them. So, it changes my point of view, I don't think that it's the best approach.

oleurud: Ahh ok, it creates a folder inside node_modules. I thought it was a `magic local link`.

So, for me, the best option is `The Global` and `The Wrapper` together. It is easy to understand and it allows to move code without problems.

I don't like other options like the environment variable, because it could cause conflict between projects.

PR #936, opened 2018-04-11
MVT: Different types/values for same property in same feature in different tiles

rochoa:
https://github.com/CartoDB/support/issues/1369


Algunenano:
This is fixed with Windshaft 4.7.1. @oleurud @rochoa Can you have a look at the changes and I'll get it merged and released?

PR #937, opened 2018-04-12
Add test to check regression when instantiating a map

dgaubert:
Add a test to check regression when cache-buster is not updated


dgaubert:
Fixed

PR #952, opened 2018-05-08
Add optional layer metadata at instantiation

### LINKED ISSUE #948: Implement Metadata as needed by CARTO VL
CartoVL is using the SQL API to obtain metadata about the dataset/query used, including a sample of the data.
We must avoid using the SQL API from CartoVL and obtain this data from the Maps API to avoid requiring both Maps API and SQL API authorization keys.

At this point we want to implement quickly what cartovl needs, and eventually refactor it into something more reasonable and efficient.

We could just implement now an ad-hoc endpoint performing the exact same queries/data processing we do now in CartoVL.

Or, if we consider the effort will be similar (I'm inclined to think so), to implement it as optional metadata returned by the map instantiation. This will offer opportunities for optimization (now or later) since some of the metadata may be already computed/used, we save requests and could also save queries by combining requested metadata with e.g. the needs of `getAggregationMetadata`).

We could add a parameter to request metadata, e.g. `"metadata": { sample: true, rowCount: true, columnStats: true }` and the data could be added to the existing `metadata.stats` in the response. This could be nicely encapsulated in the
[`setLayerStats`](https://github.com/CartoDB/Windshaft-cartodb/blob/20b46a33cf725fa8214f6c8dd9e171a14d83bb6f/lib/cartodb/controllers/map.js#L460) function of the Windshaft-cartodb maps controller.

### What CartoVL does now

All the metadata CartoVL requests now is actually needed.
Tne `windshaft` module encapsulates all SQL API requests through `getSQL`, which is used by the next functions which are called to prepare the metadata (in `_getMetadata`):

* `getSample(conf, sampling)`
  + query: `SELECT * TABLESAMPLE BERNOULLI / random() < x`
  + => `metadata.sample`
* `getFeatureCount(query, conf)`
  + query: `SELECT COUNT(*) FROM ${query}`
  + => `metadata.featureCount`
* `getColumnTypes(query, conf)`
  + query: `select * from ${query} limit 0` => .fields (name: {type})
  + => `metadata.columns`
  + => `metadata.categoryIDs, categoryIDsToName`
  + Note that, depending to the field `type`, columns are categorized as numeric, date or category (all strings).
* `getNumericTypes(names, query, conf)` (executed for numeric columns)
  + query: `SELECT min($name), max($name), ...FROM ${query}`
  + => `metadata.columns` (name, type, min, max, avg, sum)
  + `COUNT(*)` computed but not used (?)
* `getDatesTypes(names, query, conf)` (executed for date columns)
  + query:  `SELECT min($name), max($name) FROM ${query}`
  + => `metadata.columns` (name, type, min, max)
* `getCategoryTypes(names, query, conf)` (executed for category columns)
  + query: `SELECT COUNT(*), ${name} FROM ${query} GROUP BY ${name}`
  + => `metadata.categoryIDs, categoryIDsToName`
  + => `metadata.columns` (name, type, categoryNames, categoryCounts)
* `getGeometryType(query, conf)`
  + query: `SELECT ST_GeometryType(the_geom) FROM ${query}`
  + => `windshaft.geomType`

In addition to the metadata (`categoryIDs`, `columns`, `featureCount`, `sample`).
`geomType` is kept in windshaft object, used to deterimine if aggregation is possible and to decode MVT.

### What the tiler already does at instantiation

The module `query-utils` of Windshaft-cartodb contains some functions to fetch metadata about the query.
In particular a function `getAggregationMetadata` used to determined if aggregation should be applied which returns:
* An estimate of the row count
  + using `CDB_EstimateRowCount` (as `meta.stats.estimatedFeatureCount`)
* the geometry type
  + using `SELECT ST_GeometryType(${geom}) FROM (${query}) WHERE ${geom} IS NOT NULL LIMIT 1`

When default aggregation is used (sampling) the columns of the original query are obtained with a LIMIT 0 query (in `getLayerAggregationColumns`) to set the `columns` layer parameter.

The map instantiation response contains `layergroup.metadata.layers[0].meta.stats.estimatedFeatureCount` (which could be extended for additional metadata). It also contains `layergroup.metadata.layers[0].meta.stats.aggregation` which could be used for aggregated stats at some point.

jgoizueta:
This is experimental to be used by CartoVL. It shouldn't be public/documented at the moment.

Fixes #948


jgoizueta:
Notes:

This allows obtaining additional metadata in map instantiation. When aggregation occurs, the metadata is about the original, unaggregated data source.
This is so to fullfill current Carto VL needs, but we can discuss about having aggregation metadata too in the future.

The metadata can be requested adding a `metadata` entry to the layer `options`, which can have this properties:

* `featureCount`: (true/false) actual row count (as opposed to the default estimated row count)
* `geometryType` (true/false) type of the geometry column
* `columns`: (true/false) name and types of functions
* `columnStats`: (true/false/object) same as `columns` and also some basic stats: max, min, avg, sum for numbers, max, min for dates, cand top categories for text columns. The number of top categories is 1024 by default and can be controlled by passing an object like `{ topCategories: 100 }` here.
* `sample` (integer) returns a sample of the table of the approximate size (number of rows).

The requested metadata is returned in the response for each layer (`metadata.layers`) in `meta.stats`, in addition to `estimatedFeatureCount` which we already had.

I've preserved the existing behaviour for `estimatedFeatureCount`, which takes the value -1 if any SQL error occurs when computing it. I think this is questionable, as may be hiding unexpected problems. For the new metadata I'm not filtering SQL errors.

jgoizueta:
I'm reviewing this because of a major problem: at the point where metadata is being computed now only the aggregated query (in case of aggregation) is available.
We could extract the original query since it appears as `( ... ) _cdb_query` but a less hacky approach would be desirable.

Also the previously existing stat `estimatedFeatureCount` refers to the aggregated results in case of aggregation. So, to be consistent we should be explicit in the metadata parameters about pre/post aggregation metadata and support both in some cases (like count).

jgoizueta:
Well, regarding the existing stat `estimatedFeatureCount` it never really worked with aggregation because it executes the aggregation query without performing Mapnik token substitution (scale_denominator, etc.) This wasn't noticed because as stated in my previous comment we were silently returning -1 in case of error.

So, now we should be free to change the behaviour of this for aggregation and return the estimate count for the pre-aggregation query in this default stat.

jgoizueta:
Notes for code reviewing:

* `estimatedFeatureCount` is now fixed for aggregations (it used to return always -1 in that case), but note that it returns the number of features before aggregation. It still will be -1 if a SQL occurs (this is reasonable because this estimate is computed without the user asking for it)
* Other on-demand meta abput the pre-aggregation data has been added as described the comment above: `featureCount`, `geometryType`, `columns`, `columnStats`, `sample`. This are current Carto-VL needs. For these, SQL errors are transmitted, not ignored.
* A proof-of-concept aggregation metadata has been added: `aggrFeatureCount` for the number of aggregated features. This depends on the zoom level, so the value of this parameter is the zoom level for which the aggregated metadata is requested.
* We should review the way the pre-aggregation query is made available to the metadata code. Shouldn't `raw_sql` be usable for this? (it seems not).
* These changes  only affects new map instantiations that use the new metadata parameters  (except for the fix of estimatedFeatureCount with aggregations), and the new parameters aren't documented yet, so this should be save for production

jgoizueta:
I've looked at how `sql_raw` is used:

We keep the original query in `sql_raw` when `sql_wrap` is used:
https://github.com/CartoDB/Windshaft-cartodb/blob/b8109401d1bf83d723e21249d4af9c71fa13bf4f/lib/cartodb/models/mapconfig/adapter/sql-wrap-mapconfig-adapter.js#L15-L16
(in that case)

Then we use it to: 

* decide whether aggregation will be used
  + https://github.com/CartoDB/Windshaft-cartodb/blob/b8109401d1bf83d723e21249d4af9c71fa13bf4f/lib/cartodb/models/mapconfig/adapter/aggregation-mapconfig-adapter.js#L132
* generate the aggregation query from it:
  + https://github.com/CartoDB/Windshaft-cartodb/blob/b8109401d1bf83d723e21249d4af9c71fa13bf4f/lib/cartodb/models/aggregation/aggregation-mapconfig.js#L102-L103
* generate analysis queries from it:
  + https://github.com/CartoDB/Windshaft-cartodb/blob/b8109401d1bf83d723e21249d4af9c71fa13bf4f/lib/cartodb/models/mapconfig/adapter/analysis-mapconfig-adapter.js#L120-L121
* get values for Mapnik tokens (from query extents) for TurboCarto:
  + https://github.com/CartoDB/Windshaft-cartodb/blob/b8109401d1bf83d723e21249d4af9c71fa13bf4f/lib/cartodb/models/mapconfig/adapter/turbo-carto-adapter.js#L120-L121

Note that query wrapping occurs [before](Note that wrapping comes before aggregation:
https://github.com/CartoDB/Windshaft-cartodb/blob/b8109401d1bf83d723e21249d4af9c71fa13bf4f/lib/cartodb/api/api-router.js#L132-L141
) aggregation, but then we [re-wrap it](https://github.com/CartoDB/Windshaft-cartodb/blob/b8109401d1bf83d723e21249d4af9c71fa13bf4f/lib/cartodb/models/mapconfig/adapter/aggregation-mapconfig-adapter.js#L101-L104
) (aggregation query is generated using `sql_raw`, then `sql_wrap` applied after aggregation).

So I think we can use `sql_raw` to keep the query without aggregation (and unwrapped in case of wrapping).

The only thing I'm not fully sure yet is about the TurboCarto stuff;
if we have aggregations, then the SQL needs the tokens, and now, since there's no sql_raw, we must be using the aggregation query to determine the geographical extents and mapnik token values. If we change to use sql_raw for aggregations, then the original query will be used for the extent/tokens. That's probably more efficient and equivalent to using the aggregation query. So I think this should work.
(a different question is whether the way the tokens are computed makes sense for aggregation queries)

The changes to use `sql_raw` are here (I'm also removing the other `aggregation_metadata` which I kept but wasn't using yet):

```diff
diff --git a/lib/cartodb/backends/layer-stats/mapnik-layer-stats.js b/lib/cartodb/backends/layer-stats/mapnik-layer-stats.js
index 6a71435..d895d4d 100644
--- a/lib/cartodb/backends/layer-stats/mapnik-layer-stats.js
+++ b/lib/cartodb/backends/layer-stats/mapnik-layer-stats.js
@@ -275,10 +275,8 @@ function formatResultFields(dbConnection, flds) {
 
 MapnikLayerStats.prototype.getStats =
 function (layer, dbConnection, callback) {
-    let aggrQuery = layer.options.sql_raw || layer.options.sql;
-    let preQuery = layer.options.aggregation_metadata ?
-        layer.options.aggregation_metadata.pre_aggregation_sql :
-        aggrQuery;
+    let aggrQuery = layer.options.sql;
+    let preQuery =  layer.options.sql_raw || aggrQuery;
 
     let context = {
         dbConnection,
diff --git a/lib/cartodb/models/mapconfig/adapter/aggregation-mapconfig-adapter.js b/lib/cartodb/models/mapconfig/adapter/aggregation-mapconfig-adapter.js
index b3df3d9..26567ff 100644
--- a/lib/cartodb/models/mapconfig/adapter/aggregation-mapconfig-adapter.js
+++ b/lib/cartodb/models/mapconfig/adapter/aggregation-mapconfig-adapter.js
@@ -84,7 +84,7 @@ module.exports = class AggregationMapConfigAdapter {
 
     _adaptLayer (connection, mapConfig, layer, index) {
         return new Promise((resolve, reject) => {
-            this._shouldAdaptLayer(connection, mapConfig, layer, index, (err, shouldAdapt, aggrMeta) => {
+            this._shouldAdaptLayer(connection, mapConfig, layer, index, (err, shouldAdapt) => {
                 if (err) {
                     return reject(err);
                 }
@@ -102,6 +102,12 @@ module.exports = class AggregationMapConfigAdapter {
                     aggregationSql = sqlQueryWrap.replace(/<%=\s*sql\s*%>/g, aggregationSql);
                 }
 
+                if (!layer.options.sql_raw) {
+                    // if sql_wrap is present, the original query should already be
+                    // in sql_raw (with sql being the wrapped query);
+                    // otherwise we keep the now the original query in sql_raw
+                    layer.options.sql_raw = layer.options.sql;
+                }
                 layer.options.sql = aggregationSql;
 
                 mapConfig.getLayerAggregationColumns(index, (err, columns) => {
@@ -111,12 +117,6 @@ module.exports = class AggregationMapConfigAdapter {
 
                     layer.options.columns = columns;
 
-                    layer.options.aggregation_metadata = {
-                        pre_aggregation_sql: sqlQueryWrap || sqlQuery,
-                        geometry_type: aggrMeta.type,
-                        pre_aggregation_count: aggrMeta.count
-                    };
-
                     return resolve({ layer, index, adapted: shouldAdapt });
                 });
             });
@@ -161,7 +161,7 @@ module.exports = class AggregationMapConfigAdapter {
                 return callback(null, false);
             }
 
-            callback(null, true, result);
+            callback(null, true);
         });
     }
 ```

**Comment by jgoizueta** [comment]
So @dgaubert, if you have the courage to read my previous comment, do you think we should use `sql_raw` for this?

File: lib/cartodb/backends/layer-stats/mapnik-layer-stats.js [hunk 1]
```diff
@@ -11,18 +35,273 @@ MapnikLayerStats.prototype.is = function (type) {
...
+                return err;
+            }
+        )
+    );
+}
+
+function _estimatedFeatureCount(queries, ctx) {
+    if (queries.results.estimatedFeatureCount === undefined) {
+        // This is always computed; a default value of -1 is used in case of error
+        addStat(
```
dgaubert: I don't like the way that we gather results from queries here. Passing a placeholder (`queries.results`) to be populated by the function that is being called in general is not a good idea. Despite Javascript passes arguments by reference, it breaks some rules of best practices so I'd try to do the contrary:

```javascript
function queryPromise(dbConnection, query) {
    return new Promise((resolve, reject) => {
        dbConnection.query(query, function (err, res) {
            if (err) {
                return reject(err);
            }
            resolve(res);
        });
    });
}

// Each function receives the arguments that actually it needs. No need to create a
// function that receives a lot of arguments and perform one thing or another depending of them
function _estimatedFeatureCount(dbConnection, getQueryRowEstimation, query) {
    return queryPromise(dbConnection, getQueryRowEstimation(query))
        .then(res => { estimatedFeatureCount: res.rows[0].rows })
        .cath(() => -1);
}

function _aggrFeatureCount(dbConnection, getQueryActualRowCount, aggrQuery, zoom) {
    const query = queryForZoom(aggrQuery, zoom);

    return queryPromise(dbConnection, getQueryActualRowCount(query))
        .then(res => { aggrfeatureCount: res.rows[0].rows });
}

MapnikLayerStats.prototype.getStats = function (layer, dbConnection, callback) {
    let aggrQuery = layer.options.sql_raw || layer.options.sql;
    let preQuery = layer.options.aggregation_metadata ?
        layer.options.aggregation_metadata.pre_aggregation_sql :
        aggrQuery;

    // Then you can call them with the arguments that they need.
    Promise.all([
        _estimatedFeatureCount(dbConnection, queryUtils.getQueryRowEstimation, preQuery),
        _aggrFeatureCount(dbConnection, queryUtils.getQueryActualRowCount, aggrQuery)
    ])
    .then(results => {
        const resultsMerged = results.reduce((acc, current) => Object.assign(acc, current), {});
        callback(null, resultsMerged);
    })
    .catch(err => callback(err));
};
```

**Note**: To handle promises in the way that you need, see the following comment

lib/cartodb/backends/layer-stats/mapnik-layer-stats.js
```diff
@@ -11,18 +35,273 @@ MapnikLayerStats.prototype.is = function (type) {
...
+
+/* Helper to add a task to the queries PhasedExecution
+ * type can be either 'pre' (for pre-aggregation metadata) or 'post'
+ * zoom is used only for post-aggregation metadata
+ * query is a function that generates a metadata query from a data query
+ * assign is a function to assign the results of the metadata query
+ * if a assignDefault function is present, it will be used in case of error
+ * during the query execution and any errors will be ignored
+ */
+function addStat(queries, ctx, type, zoom, query, assign, assignDefault=null) {
```
dgaubert: I think you end up with something more complicated than you need actually (see other comments). This function receives a lot of parameters to cover all the cases that you have right now. If in the future we need to add more cases (get more metadata) then this function is going to get crazy.

To avoid this complexity I'd try to define a function for each different case that receives only the parameters that they need and returns a promise:

```javascript
function _estimatedFeatureCount(dbConnection, getQueryRowEstimation, preQuery) {
    ...
    return queryPromise(...).then(res => { estimatedFeatureCount: res.rows[0].rows })
}

function _aggrFeatureCount(dbConnection, getQueryActualRowCount, aggrQuery, zoom) {
    ...
    return queryPromise(...).then(res => { aggrfeatureCount: res.rows[0].rows });
}
```
Please note that these promises are resolving an object, this will be useful later when you need to merge all results from different queries `Promise.all([...]).then(results => { ... merge all results ...})`.

dgaubert:
> So @dgaubert, if you have the courage to read my previous comment, do you think we should use sql_raw for this?

As @rochoa wrote [here](https://github.com/CartoDB/product/issues/30#issuecomment-384020382): TurboCarto's ramps/legends should use the original data/query for the computation instead of the aggregated one. 

So I'd say that we should use `sql_raw`.

jgoizueta:
Regarding `sql_raw` and TurboCarto:

After reviewing with @dgaubert  the code [here](https://github.com/CartoDB/Windshaft-cartodb/blob/b8109401d1bf83d723e21249d4af9c71fa13bf4f/lib/cartodb/models/mapconfig/adapter/turbo-carto-adapter.js#L120-L121): What it does now is slightly different from what I had understood:

If sql_raw is not present, then tokens are substituted (in the non-raw query) with some generic values (e.g. the bbox is the whole Earth extents).
if sql_raw is present, then the parameters are derived from the raw query geographical extents (which become the bbox). The reason for this seems to be the hexabin, etc. queries which are used with sql_wrap and which use the bbox token. 

So, if we now also use sql_raw when aggregations are present, to avoid changing current behaviour we should add also the condition that `sql_wrap` is present to apply the data-dependent substitution. (But we could also look at what makes more sense for the aggregation queries which also uses tokens, since it is zoom-dependent)

lib/cartodb/backends/layer-stats/mapnik-layer-stats.js
```diff
@@ -11,18 +35,273 @@ MapnikLayerStats.prototype.is = function (type) {
...
+    let sql;
+    if (type === 'pre') {
+        sql = ctx. preQuery;
```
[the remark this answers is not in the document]
jgoizueta: oopsie!

jgoizueta:
Hey @dgaubert can you review this again?

There was an interesting problem with the tests (well, "interesting" wasn't the word I had in mind while I was pulling my hair out trying to figure it out).

The test "layergroup can hold substitution tokens" was failing but only on travis. (not even using the docker tests locally). I finally was able to reproduce it locally by setting `global.environment.enabledFeatures.layerStats = true;` in the test.
So we have some test-order issue with some test leaving an indeterminate state in `layerStats`.

Now, the problem, which has existed for a long time, and has been reveled now, is this: we always compute the row count estimate stat. But this has been failing if the sql query contains Mapnik tokens (because we make no substitution before executing it).
But since we were ignoring any SQL errors for that query (setting row count to -1) we didn't notice.

I've fixed the substitution problem, but I haven't look at the `layerStats` configuration problem

jgoizueta:
I've reverted the behaviour in case of error when computing `estimatedFeatureCount` (computed in each instantiation) to ignore errors and return -1, so we can safely deploy this in production without introducing new risks.

@oleurud do you think worth to make that conditional on the environment (so that in development, staging, etc the errors aren't ignored)? (`if (process.env.NODE_ENV === 'production') ...`)

oleurud:
Maybe an environment configuration parameter will be the best option (easy and fast to enable/disable)

lib/cartodb/backends/layer-stats/mapnik-layer-stats.js
```diff
@@ -11,17 +34,280 @@ MapnikLayerStats.prototype.is = function (type) {
...
+    // (if metaOptions.geometryType) from it.
+
+    // TODO: compute _sample with _featureCount when available
+
+    Promise.all([
+        _estimatedFeatureCount(ctx).then(
+            ({ estimatedFeatureCount }) => _sample(ctx, estimatedFeatureCount)
+                .then(s => mergeResults([s, { estimatedFeatureCount }]))
+        ),
+        _featureCount(ctx),
```
oleurud: A question to understand it: If some of _featureCount, _aggrFeatureCount, _geometryType or _columns fails, we will return an error as a response. I know that the user must request for it expressly, but I am not sure if metadata should do to fail a request. What is your point of view?

lib/cartodb/backends/layer-stats/mapnik-layer-stats.js
```diff
@@ -11,17 +34,279 @@ MapnikLayerStats.prototype.is = function (type) {
...
+        });
+
+    });
+}
+
+function columnAggregations(field) {
+    if (field.type === 'number') {
+        return ['min', 'max', 'avg', 'sum'];
+    }
+    if (field.type === 'date') { // TODO other types too?
```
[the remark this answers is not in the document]
oleurud: Forget if, you are right.

lib/cartodb/backends/layer-stats/mapnik-layer-stats.js
oleurud: It is not a reason to stop the PR and also, I am not saying to change it. Only it raises doubts.

test/acceptance/aggregation.js
```diff
@@ -2173,6 +2187,168 @@ describe('aggregation', function () {
...
+                                type: 'cartodb',
+
+                                options: {
+                                    sql: POINTS_SQL_PAIRS,
+                                    resolution: 1,
+                                    aggregation: {
+                                        threshold: 1
+                                    },
+                                    metadata: {
+                                        aggrFeatureCount: 10
```
dgaubert: 10 is the zoom value??

jgoizueta: Exactly

lib/cartodb/backends/layer-stats/mapnik-layer-stats.js
dgaubert: IMHO, If the user requests for any specific metadata and an error happens, we should return that error because we weren't able to process the request completely.

lib/cartodb/backends/layer-stats/mapnik-layer-stats.js
```diff
@@ -11,17 +34,278 @@ MapnikLayerStats.prototype.is = function (type) {
...
+    // TODO: could save some queries if queryUtils.getAggregationMetadata() has been used and kept somewhere
+    // we would set queries.results.estimatedFeatureCount and queries.results.geometryType
+    // (if metaOptions.geometryType) from it.
+
+    // TODO: compute _sample with _featureCount when available
+
+    Promise.all([
+        _estimatedFeatureCount(ctx).then(
+            ({ estimatedFeatureCount }) => _sample(ctx, estimatedFeatureCount)
+                .then(s => mergeResults([s, { estimatedFeatureCount }]))
```
dgaubert: 🙏 rename `s` by `sample`

lib/cartodb/backends/layer-stats/mapnik-layer-stats.js
```diff
@@ -11,17 +34,278 @@ MapnikLayerStats.prototype.is = function (type) {
     return this._types[type] ? this._types[type] : false;
 };
 
+function queryPromise(dbConnection, query, adaptResults, errorHandler) {
```
dgaubert: Isn't it a little bit of complication here?

We can simply return a promise:
```javascript
function queryPromise(dbConnection, query) {
  return new Promise((resolve, reject) => { 
    dbConnection.query(query, (err, res) => err ? reject(err) : resolve(res))   
  })
}
```
And then, for each specific metadata function:
```javascript
function _estimatedFeatureCount (ctx) {
  return queryPromise(ctx.dbConnection, _getSQL(ctx, queryUtils.getQueryRowEstimation))
    .then(res => ({ estimatedFeatureCount: res.rows[0].rows }))
    .catch(() => ({ estimatedFeatureCount: -1 }))
}
```
Note: Remember, `.then` and `.catch` return a promise as well.

dgaubert: Indeed, we can avoid the `ctx` object and pass only the arguments that each metadata function needs.

jgoizueta: I have fixed the other issues you mentioned (except the queryPromise refactor), as they were very simple.

Regarding queryPromise, I think you're absolutely right, but I don't want to delay the deploy today, so I'll open a separate PR to do it.

dgaubert: Fair enough! 

Please, don't miss that refactor; you'll end up loving promises and their composability!

💋

Issue #771, opened 2017-10-10
Investigate pg_error in production while fetching dataview.


dgaubert:
From logs:
```
Uncaught exception: TypeError: Cannot read property 'pg_typeof' of undefined
    at psql.query (/home/ubuntu/www/node-windshaft/releases/20171004151357/lib/cartodb/models/dataview/base.js:55:42)
    at Query.<anonymous> (/home/ubuntu/www/node-windshaft/releases/20171004151357/node_modules/cartodb-psql/lib/psql.js:302:24)
    at Query.handleReadyForQuery (/home/ubuntu/www/node-windshaft/releases/20171004151357/node_modules/pg/lib/query.js:124:10)
    at Connection.<anonymous> (/home/ubuntu/www/node-windshaft/releases/20171004151357/node_modules/pg/lib/client.js:176:19)
    at emitOne (events.js:101:20)
    at Connection.emit (events.js:188:7)
    at Socket.<anonymous> (/home/ubuntu/www/node-windshaft/releases/20171004151357/node_modules/pg/lib/connection.js:138:12)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:176:18)
```

It's related to histogram trying to discover column type in order to select time-series or histogram

oleurud:
The error shows the row not exists. I have created a histogram widget using the column X and then I have deleted this column X in the data section of the layer. As a result, I have this error.

Maybe the solution for this problem, it is to check the columns used by the widgets before deleting it

oleurud:
The case explained in my previous comment is right, but it is not the normal use case.

The problem usually is: in order to get the typeof of a column, we use a filtered query and these filters could have "removed" data from the column of from every row and we could have the problem. So the idea is to use all the initial data to extract the typeof (use the query without filters)

oleurud:
We are using always the queries without filters for every dataview type.

I can't reproduce the error doing "rational" things in builder. In my opinion, the error happens when someone creates a map with dataviews, then modify the dataset removing the column or updating his name and as a result, the error raises.

If it is really the case, I think that the API must return an error warning about the problem.

oleurud:
After talking with @dgaubert, the main objective is to catch the error. 

The error happens in a method of the base dataview class (all the dataviews extends from it). The method is used by the rest of dataviews and everyone check and set a default type if something was wrong.

The code is ready to CR

PR #953, opened 2018-05-10
handling pg_typeof == undefined error in dataviews

oleurud:
Fix: https://github.com/CartoDB/Windshaft-cartodb/issues/771


test/acceptance/dataviews/error-cases.js
```diff
@@ -3,85 +3,211 @@ require('../../support/test_helper');
 var assert = require('../../support/assert');
 var TestClient = require('../../support/test-client');
 
-describe('histogram-dataview', function() {
+describe('dataview error cases', function() {
```
dgaubert: I'm missing a test that actually checks the error that we are throwing here.

oleurud: ----> https://github.com/CartoDB/Windshaft-cartodb/issues/771#issuecomment-388111113

So, there is no error, the dataview uses the default value

dgaubert: You are right! Sorry for the noise

PR #954, opened 2018-05-11
Controllers goodies

dgaubert:
Apply same interfaz for all controllers:

- `.register()`: mounts the specified middleware function or functions at the specified path.
- `.middlewares()`: build and return a set of middlewares to be mounted

Also, use ES6 class syntax for some of them.

### DIFF
--- a/lib/cartodb/api/template/admin-template-controller.js
+++ b/lib/cartodb/api/template/admin-template-controller.js
@@ -3,70 +3,90 @@ const credentials = require('../middlewares/credentials');
 const rateLimit = require('../middlewares/rate-limit');
 const { RATE_LIMIT_ENDPOINTS_GROUPS } = rateLimit;
 
-/**
- * @param {AuthBackend} authBackend
- * @param {PgConnection} pgConnection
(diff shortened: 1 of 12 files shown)


lib/cartodb/api/template/admin-template-controller.js
```diff
@@ -3,70 +3,90 @@ const credentials = require('../middlewares/credentials');
...
-        credentials(),
-        authorizedByAPIKey({ authBackend: this.authBackend, action: 'list', label: 'GET TEMPLATE LIST' }),
-        rateLimit(this.userLimitsBackend, RATE_LIMIT_ENDPOINTS_GROUPS.NAMED_LIST),
-        listTemplates({ templateMaps: this.templateMaps })
-    );
+        return [
+            credentials(),
+            authorizedByAPIKey({ authBackend: this.authBackend, action, label }),
+            rateLimit(this.userLimitsBackend, rateLimitGroup),
+            checkContentType({ action: 'POST', label: 'POST TEMPLATE' }),
```
oleurud: You could remove the `action` parameter

PR #956, opened 2018-05-14
Express Routers: Activate mergeParams option

dgaubert:
In order to extract common middlewares between routers and controllers we need to activate `mergeParams` option to preserve the `req.params` from the parent router


Algunenano:
Fixes it for me. Having a test for the future would be :top:

dgaubert:
@Algunenano 👀 => https://github.com/CartoDB/Windshaft-cartodb/pull/956/commits/63a9d58c6713c065818c09813007c5d87cee7ac0

PR #962, opened 2018-05-21
Upgrade dependencies

dgaubert:


dgaubert:
> I understand that the dev branch will go away before merging.

I performed a little test on our servers to make sure that we don't break anything before publishing a new version.

PR #963, opened 2018-05-21
Refactor metadata queryPromise 

jgoizueta:
This is to fix a couple of pending issues from https://github.com/CartoDB/Windshaft-cartodb/pull/952

- [x] Remove unnecessary complexity from `queryPromise` (aka as finally fully embrace Promises)
- [ ] Replace `ctx` context parameter in metadata query functions by individual needed parameters

I have doubts about the latter: mostly about what to do with the `preQuery`, `aggrQuery`. Now each metadata function decides wether to use `post` or `pre` aggregation querying and lets the `_getSQL` helper to handle the SQL. Maybe the caller to the metadata functions should decide if it's post/pre and pass only the needed query. I need to rethink, this.


jgoizueta:
Hey @dgaubert can you CR this, so we merge it as is? I'll look at the context parameter in a separate ticket because I need more time for it.

PR #964, opened 2018-05-22
Add metadata parameter for NULL categories

jgoizueta:
This is a small changed we need to control whether ignoring NULLs or not for metadata categories information.


jgoizueta:
I don't know what's wrong with the tests 😓 They seem unstable: I restarted them and they failed differently.

PR #968, opened 2018-05-28
Sample columns

jgoizueta:
Add parameters to select sample columns

This changes the `sample` metadata parameter which was an integer (number of rows) to and object with `num_rows` (for the previous value) and `include_columns` for the sample columns.

This is relevant because for example in CARTO VL geometry is not needed in the sample and in some cases it could be huge.

I'm adding an additional  `exclude_columns` option to the sample parameter, which is WIP, but we're interesting in deploying this as is to avoid future breaking changes for CARTO VL (I mean avoid new tiler versions to break released versions of CARTO VL).


jgoizueta:
#963 should be merged first; this is based on that branch

PR #971, opened 2018-05-31
Send dates as unix epoch instead strings in .mvt files

IagoLast:
https://github.com/CartoDB/carto-vl/issues/130
--- 

## Pending

- [x] Add a parameter to control the date wrapping
- [x] Better code quality
- [x] Add tests

## summary
This creates a new ConfigAdapter used in vector maps instantiations.

This adapter generates a new sql query where the date columns are wrapped into a epoch using the `date_part` function.

Doing this we make`mvt`files are smaller since we use numbers instead strings to represent dates. This is also faster in carto-vl where we interpolate dates between 0 and 1 to create animations.

We activate this feature using a `dates_as_numbers` parameter in the map instantiation.

Dates come in **SECONDS**


IagoLast:
@jgoizueta @elenatorro @danicarrion Ready to CR  👀

danicarrion:
assuming it was @dgaubert  who should be mentioned here :-)

lib/cartodb/models/mapconfig/adapter/vector-mapconfig-adapter.js
```diff
@@ -0,0 +1,61 @@
...
+            });
+    }
+
+    _getColumns(user, originalQuery) {
+        return new Promise((resolve, reject) => {
+            this.pgConnection.getConnection(user, (err, connection) => {
+                if (err) {
+                    return reject(err);
+                }
+                const query = `SELECT * FROM (${originalQuery}) _cdb_column_type limit 0`;
```
jgoizueta: There's a `getQueryLimited(sql, limit)` function in query_utils you can use here

lib/cartodb/models/mapconfig/adapter/vector-mapconfig-adapter.js
```diff
@@ -0,0 +1,61 @@
...
+            return callback(new Error('Get column types for multiple vector layers is not implemented'));
+        }
+
+        this._wrapDates(requestMapConfig, user)
+            .then(updatedRequestMapConfig => callback(null, updatedRequestMapConfig))
+            .catch(callback);
+    }
+
+    _wrapDates(requestMapConfig, user) {
+        const originalQuery = requestMapConfig.layers[0].options.sql;
```
jgoizueta: This should be applied to all layers that have `dates_as_numbers` in its options.

jgoizueta: Oh, I see your comment above now. I think we should do this with a per-layer option right now

lib/cartodb/utils/date-wrapper.js
```diff
@@ -0,0 +1,39 @@
...
+function _isDateType(field) {
+    return DATE_OIDS.hasOwnProperty(field.dataTypeID);
+}
+
+/**
+ * Return a sql query to transform a date column into a unix epoch
+ * @param {string} column - The name of the date column
+ */
+function _castColumnToEpoch(columnName) {
+    return `date_part('epoch', ${columnName}) as ${columnName}`;
```
jgoizueta: I'd add double quotes to the column names (`date_part('epoch', "${columnName}") as "${columnName}"` to avoid problems with special characters in column names.

lib/cartodb/utils/query-utils.js
```diff
@@ -162,10 +162,19 @@ module.exports.getQueryGeometryType = function(query, geometryColumn) {
     `;
 };
 
-module.exports.getQueryLimited = function(query, limit=0) {
+function getQueryLimited(query, limit = 0) {
```
jgoizueta: Mmm, so did you avoid using this in the wrapping for some reason?

lib/cartodb/utils/query-utils.js
```diff
@@ -162,10 +162,19 @@ module.exports.getQueryGeometryType = function(query, geometryColumn) {
...
+}
+
+function queryPromise(dbConnection, query) {
+    return new Promise((resolve, reject) => {
+      dbConnection.query(query, (err, res) => err ? reject(err) : resolve(res));
+    });
+}
+
+module.exports.queryPromise = queryPromise;
+module.exports.getQueryLimited = getQueryLimited;
```
jgoizueta: Ah, well, it's exported after all 😛

lib/cartodb/models/mapconfig/adapter/vector-mapconfig-adapter.js
```diff
@@ -0,0 +1,61 @@
...
+ *  - This middleware is ONLY activated when the `dates_as_numbers` option is enabled for some layer in the mapConfig.
+ *  - TODO: We currently support one single layer and we should define what to do with multiple layers.
+ */
+class VectorMapConfigAdapter {
+    constructor(pgConnection) {
+        this.pgConnection = pgConnection;
+    }
+
+    getMapConfig(user, requestMapConfig, params, context, callback) {
+        if (!this._isDatesAsNumbersFlagEnabled(requestMapConfig)) {
```
dgaubert: As mentioned in other comments, this should be checked at layer level and adapt the query in case the flag is enabled.

lib/cartodb/models/mapconfig/adapter/vector-mapconfig-adapter.js
```diff
@@ -0,0 +1,61 @@
...
+            return callback(null, requestMapConfig);
+        }
+
+        if (requestMapConfig.layers.lenght > 1) {
```
dgaubert: CartoVL is just one client of this service, we need to support multiple vector layers to be ready for other clients.

Algunenano: lenght > length

lib/cartodb/models/mapconfig/adapter/vector-mapconfig-adapter.js
IagoLast: fixed

test/acceptance/date-wrapping.spec.js
```diff
@@ -38,6 +38,18 @@ describe('date-wrapping', () => {
...
                 });
             });
+
+            it('should return metadata with casted columns', done => {
+
+                testClient.getLayergroup(function(err, layergroup) {
+                    const expected = ['date'];
+                    assert.ifError(err);
+                    assert.deepEqual(layergroup.metadata.layers[0].meta.dates_as_numbers.sort(), expected.sort());
+                    testClient.drain(done);
```
IagoLast: this is not needed (already in the `afterEach` hook)

jgoizueta:
@dgaubert can you re-check this? @IagoLast fixed it and I added metadata when the date/integer wrapping is performed which contains the names of the columns that were converted. (this allows CARTO VL or any other client to know which numeric columns contains timestamps)

test/acceptance/date-wrapping.spec.js
```diff
@@ -0,0 +1,109 @@
...
+
+
+    describe('when a map instantiation has multiple layers', () => {
+        beforeEach(() => {
+            const mapConfig = mapConfigFactory.getVectorMapConfig({ numberOfLayers: 2 });
+            testClient = new TestClient(mapConfig);
+        });
+        describe('and both layers have the "dates_as_numbers" option enabled', () => {
+            // TODO: Pending test
+            it('should return dates as numbers for every layer');
```
dgaubert: Sorry, this must be achieved before proceeding.

test/acceptance/date-wrapping.spec.js
```diff
@@ -0,0 +1,109 @@
...
+        });
+
+        describe('and only one layers has the "dates_as_numbers" option enabled', () => {
+            // TODO: Pending test
+            it('should return dates as numbers only for the layer with the "dates_as_numbers" flag enabled');
```
dgaubert: Sorry, this must be achieved before proceeding.

test/acceptance/date-wrapping.spec.js
```diff
@@ -0,0 +1,109 @@
...
+        });
+
+        describe('and none of the layers has the "dates_as_numbers" option enabled', () => {
+            // TODO: Pending test
+            it('should return dates as dates for both layers');
```
dgaubert: Sorry, this must be achieved before proceeding.

dgaubert:
🙏  Fix broken test!!!

lib/cartodb/utils/date-wrapper.js
```diff
@@ -35,17 +35,19 @@ function _castColumnToEpoch(columnName) {
 }
 
 function wrappedDates(query) {
```
IagoLast: IMO `getColumnsWithWrappedDates` would be the perfect name

jgoizueta:
Hey @dgaubert we have the tests and nothing seems broken 😅 Do you want to take another look?

lib/cartodb/utils/date-wrapper.js
jgoizueta: OK, chaning it...

Algunenano:
Can you also add or update the docs, please?

jgoizueta:
I had just put an "experimental" note in the NEWS, @Algunenano, because we don't want to document this yet: we'll be using it in CARTO VL and we might want to make breaking changes based on that experience.

We'll accumulating undocumented features to support CARTO VL for some time, and we should eventually make it all public and document it, when we're confident the API is OK.

PR #973, opened 2018-06-06
Accept `layer` option in static map url request

Algunenano:
- Fixes https://github.com/CartoDB/cartodb/issues/13865 by accepting the `layer` in the url as done for the named maps.
- Increases the threshold for the static map image test.
- Removes direct usage of `node-postgres` in the tests and uses `cartodb-psql` instead. For me this fixes some failures the torque timeout tests.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -39,6 +39,7 @@ Bug Fixes:
 - Static maps fails for unsupported formats
 - Handling errors extracting the column type on dataviews
 - Fix `meta.stats.estimatedFeatureCount` for aggregations and queries with tokens
+- Static maps filters correctly if `layer` option is passed in the url.
 
 ## 6.1.0
 Released 2018-04-16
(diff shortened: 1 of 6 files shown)


test/acceptance/named_maps_static_view.js
```diff
@@ -19,8 +19,8 @@ describe('named maps static view', function() {
     var username = 'localhost';
     var templateName = 'template_with_view';
 
-    var PNG_IMAGE_TOLERANCE = 20;
-    var JPG_IMAGE_TOLERANCE = 100;
```
enekid: Why is this increase necessary?

Algunenano: Subpixel precision in these images. This works fine under travis but since the result depends on whether you have the marker symbolizer cache enabled, if you are using the static build we provide for Mapnik (or other) and what compilation flags are being used it was complaining in several local environments (mine and @dgaubert at least).

dgaubert: I'd keep these values as they are. When I ran into this problem I had an issue with my configuration (`cache-features` set to true in the test environment). In your case, you have a very custom environment, having in mind that, you can live with that by skipping these tests just for you.

Algunenano: :ok_hand: I'll remove that commit from the PR

PR #976, opened 2018-06-11
Test: Make POSTGIS_VERSION detection automatic

Algunenano:
Same idea as https://github.com/CartoDB/Windshaft/pull/627

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -135,7 +135,7 @@ fi
 
 echo "Preparing the environment"
 cd ${BASEDIR}/test/support
-sh prepare_db.sh ${PREPARE_DB_OPTS} || die "database preparation failure"
+source prepare_db.sh ${PREPARE_DB_OPTS} || die "database preparation failure"
 cd -
 
 PATH=node_modules/.bin/:$PATH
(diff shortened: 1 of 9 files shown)


run_tests.sh
```diff
@@ -5,7 +5,7 @@ OPT_CREATE_PGSQL=yes # create the PostgreSQL test environment
 OPT_DROP_REDIS=yes   # drop the redis test environment
 OPT_DROP_PGSQL=yes   # drop the PostgreSQL test environment
 OPT_COVERAGE=no      # run tests with coverage
-OPT_DOWNLOAD_SQL=yes # download a fresh copy of sql files
+OPT_DOWNLOAD_SQL=no  # download a fresh copy of sql files
```
dgaubert: Travis will need this set to `yes` to get sql functions.

enekid: Travis is failing. Could it be because of this?

oleurud: :+1:

Algunenano: Yeah, I shouldn't have push that change

Algunenano: Yeah, I shouldn't have pushed that change. I'm removing it now

Algunenano:
> Apart from detecting PostGIS Version automatically, what do you want to achieve with this PR?

For this PR just that, basically so if you are running the tests locally and you have a compatible Postgis release all ST_AsMVT tests are run too.

PR #989, opened 2018-07-03
Upgrading cartodb-redis

oleurud:
Updating `cartodb-redis` package:
- fix bad behavior with an endpoint with several rate limits. More info: https://github.com/CartoDB/node-cartodb-redis/pull/26
- new major version removing deprecated functions https://github.com/CartoDB/node-cartodb-redis/issues/28


oleurud:
It was already ready to use the new version (it is not using the deprecated functions)

PR #992, opened 2018-07-06
Fix some rate limit tests

rafatower:
`rateLimitsByEndpoint` is defined in the test config, but usually
there's no guarantee it is defined in the development config. This makes
the tests more robust.

### DIFF
--- a/test/acceptance/rate-limit.test.js
+++ b/test/acceptance/rate-limit.test.js
@@ -160,7 +160,9 @@ function assertRateLimitRequest (status, limit, remaining, reset, retry, done) {
 describe('rate limit', function() {
     before(function() {
         global.environment.enabledFeatures.rateLimitsEnabled = true;
-        global.environment.enabledFeatures.rateLimitsByEndpoint.anonymous = true;
+        global.environment.enabledFeatures.rateLimitsByEndpoint = {
+            anonymous: true
+        };
 
         redisClient = redis.createClient(global.environment.redis.port);
         testClient = new TestClient(createMapConfig(), 1234);


test/acceptance/rate-limit.test.js
```diff
@@ -160,7 +160,9 @@ function assertRateLimitRequest (status, limit, remaining, reset, retry, done) {
 describe('rate limit', function() {
     before(function() {
         global.environment.enabledFeatures.rateLimitsEnabled = true;
-        global.environment.enabledFeatures.rateLimitsByEndpoint.anonymous = true;
+        global.environment.enabledFeatures.rateLimitsByEndpoint = {
```
oleurud: Why does it make more robust? 

With this change, your are overwriting all the endpoints values https://github.com/CartoDB/Windshaft-cartodb/blob/a6644f59f7134ef363955038812d7307a4fb7ca2/config/environments/test.js.example#L370 removing all the endpoints less the `anonymous` one. Really, although is a global change for every test, it does not change the behaviour in the rest because the endpoints are setted as `false` as default and now will be `undefined`

rafatower: Otherwise it fails with:
```
  1) rate limit "before all" hook:
     TypeError: Cannot set property 'anonymous' of undefined
      at Context.<anonymous> (test/acceptance/rate-limit.test.js:163:75)
[...]
```
you don't want your test to depend on someone's config if you can avoid it. Call it developer happiness

dgaubert: By removing your `config/environments/test.js` and running the tests again your `test.js` will be created from `test.js.example`

oleurud: I see the problem now, but the solution must be set the default values even if they are not set.  Let me check

rafatower:
`rateLimitsByEndpoint` is defined in:
```
$ git grep  -l rateLimitsByEndpoint -- config/
config/environments/development.js.example
config/environments/production.js.example
config/environments/staging.js.example
config/environments/test.js.example
```

but not in my `config/environments/development.js`. That is why those tests fail with the `Cannot set property 'xyz' of undefined` (that `undefined` refers to `enabledFeatures.rateLimitsByEndpoint`).

So, it is **my fault** if the tests do not work in **my environment**. If you feel this PR makes the tests worse, let''s just discard it. I won't take it personally :slightly_smiling_face: If you feel it can be improved, go for it! either way, I'll fix my local env.

My personal feeling: I wouldn't make tests config dependent, but... :man_shrugging:

test/acceptance/rate-limit.test.js
rafatower: > By removing your config/environments/test.js and running the tests again your test.js will be created from test.js.example

uhm, did not now that. How about renaming `test.js.example` to `test.js`?

rafatower: quick question how is the environment (development/test/production) chosen when running tests?

oleurud: As a parameter when we start the app: 
`node app.js XXXX`
being XXXX == development | test | ...

oleurud:
We had a related issue https://github.com/CartoDB/CartoDB-SQL-API/issues/484 As you can see, I said it is not a problem for `windshaft-carto` due to it wasn't a production problem if the configuration files were not well defined, because the app works fine. 

For the development/test cases, the idea is "the dev must know". But I am not totally sure if it is the best option, because we can lose time finding what is the problem...

> My personal feeling: I wouldn't make tests config dependent, but... 

I agree as a global idea (totally in unit tests), but the app is based on environment configuration file. So, especially in acceptance tests, if we don't use the configuration file in tests:
- we must add the configuration params in a different way (different from production way)
- and we must do it for every test
The idea is to have a default behavior as a base, and then, change it if needed in every test.

test/acceptance/rate-limit.test.js
dgaubert: > uhm, did not now that. How about renaming test.js.example to test.js?

It's what the test runner does at the very beginning.

dgaubert: > quick question how is the environment (development/test/production) chosen when running tests?

[Here](https://github.com/CartoDB/Windshaft-cartodb/blob/028c3f9aeca6f39ad8736fc7d46724b51410d2ab/test/support/test_helper.js#L21) while testing and using [process argument](https://github.com/CartoDB/Windshaft-cartodb/blob/028c3f9aeca6f39ad8736fc7d46724b51410d2ab/app.js#L36), for instance `node app.js development`.  when you want to run the service.

PR #1006, opened 2018-07-17
Upgrade windshaft@4.8.2

dgaubert:


dgaubert:
Good catch, I've forgot totally

enekid:
It happens to me all the time xD

PR #1012, opened 2018-07-30
Remove step filter stats

oleurud:


lib/cartodb/backends/filter-stats.js
```diff
@@ -24,36 +23,30 @@ function getEstimatedRows(pgQueryRunner, username, query, callback) {
...
+            getEstimatedRows(this.pgQueryRunner, username, query, (err, rows) => {
+                if (err){
+                    return callback(err);
+                }
+
+                stats.filtered_rows = rows;
+                return callback(null, stats);
+            });
+
+        } else {
```
dgaubert: Please, user `early returns` instead of `if/else` statements

oleurud: :+1:

PR #1013, opened 2018-07-30
Remove step user limits

oleurud:


lib/cartodb/backends/user-limits.js
```diff
@@ -41,44 +39,37 @@ UserLimitsBackend.prototype.getRenderLimits = function (username, apiKey, callba
...
-                if (err) {
-                    return next(err);
-                }
-
-                return next(null, userApiKey === apiKey);
-            });
-        },
-        function getUserTimeoutRenderLimits(err, authorized) {
-            var next = this;
+            callback(
```
enekid: `return callback(`

dgaubert: Yep, I ended up for being as much explict as possible. With explicit return, there is no temptation to add more code after it.

lib/cartodb/backends/user-limits.js
```diff
@@ -41,44 +39,37 @@ UserLimitsBackend.prototype.getRenderLimits = function (username, apiKey, callba
 };
 
 UserLimitsBackend.prototype.getTimeoutRenderLimit = function (username, apiKey, callback) {
-    var self = this;
-
-    step(
-        function isAuthorized() {
-            var next = this;
+    isAuthorized(this.metadataBackend, username, apiKey, (err, authorized) => {
```
dgaubert: I'm having mixed feeling with this function, which improves readability but doesn't do so much and receives four params... Do you have more pros to keep this function?

oleurud: In the function, we have 2 success calls to callback: if `!apiKey` and after checking the `apikey` against redis value asynchronously. The function allow us to solve the problem without repeat code. Other option could be to extract to content of `getTimeoutRenderLimit` to a function and call it in the sync and async previous cases

dgaubert: Ok, I see your point. Fair enough.

PR #1019, opened 2018-08-14
create carto-package.json file

oleurud:
One of https://github.com/CartoDB/onpremises/issues/575

### DIFF
--- a/carto-package.json
+++ b/carto-package.json
@@ -1,0 +1,9 @@
+{
+    "name": "carto_windshaft",
+    "current_version": {
+        "requires": {
+        },
+        "works_with": {
+        }
+    }
+}
--- a/carto-package.json
+++ b/carto-package.json
@@ -2,15 +2,15 @@
     "name": "carto_windshaft",
     "current_version": {
         "requires": {
-            "node": ">=6.9.2",
+            "node": ">=6.9.2 <10.0",
             "yarn": ">=0.27.5 <1.0.0"
         },
         "works_with": {
             "redis": ">=3.0.0",
-            "postgresql": "9.5.2",
-            "postgis": "2.2.2.0",
-            "mapnik": "",
-            "gdal":""
+            "postgresql": ">=9.5",
+            "postgis": ">=2.2",
+            "mapnik": ">=3.0.12  checking if needed",
+            "gdal":">=2.1.1  checking if needed"
         }
     }
 }
--- a/carto-package.json
+++ b/carto-package.json
@@ -2,8 +2,15 @@
     "name": "carto_windshaft",
     "current_version": {
         "requires": {
+            "node": ">=6.9.2",
+            "yarn": ">=0.27.5 <1.0.0"
         },
         "works_with": {
+            "redis": ">=3.0.0",
+            "postgresql": "9.5.2",
+            "postgis": "2.2.2.0",
+            "mapnik": "",
+            "gdal":""
         }
     }
 }
--- a/carto-package.json
+++ b/carto-package.json
@@ -9,8 +9,7 @@
             "redis": ">=3.0.0",
             "postgresql": ">=9.5",
             "postgis": ">=2.2",
-            "mapnik": ">=3.0.12  checking if needed",
-            "gdal":">=2.1.1  checking if needed"
+            "mapnik": ">=3.0.15"
         }
     }
 }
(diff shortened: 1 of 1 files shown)


carto-package.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "carto_windshaft",
+    "current_version": {
+        "requires": {
+            "node": ">=6.9.2",
+            "yarn": ">=0.27.5 <1.0.0"
+        },
+        "works_with": {
+            "redis": ">=3.0.0",
+            "postgresql": "9.5.2",
```
rafatower: that one should be
```
"postgres": ">=9.5"
```

carto-package.json
```diff
@@ -0,0 +1,16 @@
...
+            "postgis": "2.2.2.0",
```
rafatower:
```
"postgis": ">=2.2"
```

carto-package.json
```diff
@@ -0,0 +1,16 @@
...
+            "mapnik": "",
+            "gdal":""
```
rafatower: what is `gdal` needed for?

carto-package.json
```diff
@@ -0,0 +1,16 @@
...
+    "current_version": {
+        "requires": {
+            "node": ">=6.9.2",
+            "yarn": ">=0.27.5 <1.0.0"
+        },
+        "works_with": {
+            "redis": ">=3.0.0",
+            "postgresql": "9.5.2",
+            "postgis": "2.2.2.0",
+            "mapnik": "",
```
rafatower: I think this shall be `">=3.0.12" ` or `">=3.0.15" but I'm not entirely sure. @Algunenano?

rafatower: Oh, and in this case it may be legit to move it to the "requires" section, right?

carto-package.json
rafatower: I think it is used for the SQL API (export) but not for Winshaft.

carto-package.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "carto_windshaft",
+    "current_version": {
+        "requires": {
+            "node": ">=6.9.2",
```
Algunenano: For now I'd also set a max: `">=6.9.2 < 10.0"`

carto-package.json
oleurud: I have created the same initial file for both and I have forgotten to remove `gdal` from Windshaft and `mapnik` from SQL

Thank you for your comments

Algunenano: It's used in a similar way as mapnik. Windshaft-cartodb -> Windshaft -> millstone -> srs -> node-gdal -> gdal

It has a peculiarity, unless `--shared-gdal` is passed it will download Gdal 2.1.1 and compile it, otherwise it will use whatever you have installed in your system

carto-package.json
skgsergio: We don't have/use prebuilts in onprem, we build from source so Mapnik should be specified here in this case. Regarding gdal, not sure if windshaft uses it directly, if it does then yes it should be defined, if not we can remove it I think.

oleurud: Ok @skgsergio. Other question: should we add libs as `libcairo2`?

I will set the `">=3.0.15"` for mapnik, ok @Algunenano?

Thanks all :)

Algunenano: Since Windshaft relies on having metrics available please set `>=3.0.15.7`. We could also consider the latest (`3.0.15.9`) but there isn't anything forcing us to do it right now.

carto-package.json
oleurud: From https://github.com/naturalatlas/node-gdal: 
```
GDAL 2.0.1 (GEOS 3.4.2, Proj.4 4.8.0) comes bundled, so node-gdal will work straight out of the box
[...]
To link against shared libgdal, install using:
```
```sh
# requires libgdal-dev (debian: sudo apt-get install libgdal-dev)
$ npm install gdal --build-from-source --shared_gdal
```
We dont make anything especial on the installation, so, I think we dont need to add gdal as a dep

oleurud:
@lekum @skgsergio 

For the record, it also needs: libcairo2-dev, libpango1.0-dev, libjpeg8-dev and libgif-dev for server side canvas support (not sure about libgif-dev).

PR #1023, opened 2018-08-24
Document named and static maps

dgaubert:
Document named maps and static maps into the reference


inigomedina:
This looks great. 👍

PR #1024, opened 2018-08-29
Log overviews usage

enekid:
Comes from https://github.com/CartoDB/cartodb-management/issues/5201

### DIFF
--- a/lib/cartodb/models/dataview/base.js
+++ b/lib/cartodb/models/dataview/base.js
@@ -39,6 +39,9 @@ module.exports = class BaseDataview {
 
                 if (flags && flags.usesOverviews) {
                     stats.usesOverviews = true;
+                    if (this.getType) {
+                        stats.dataviewType = this.getType();
+                    }
                 }
                 
                 return callback(null, result, stats);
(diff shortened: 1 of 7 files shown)


enekid:
The dataviews part is already completed.

lib/cartodb/models/dataview/base.js
```diff
@@ -34,8 +34,14 @@ module.exports = class BaseDataview {
                 result = this.format(result, override);
                 result.type = this.getType();
 
-                return callback(null, result);
+                //Overviews logging
+                const stats = {};
 
+                if (flags && flags.usesOverviews) {
+                    stats.usesOverviews = true;
```
dgaubert: Shouldn't it be initialized to `false`?

enekid: `flags.usesOverviews` is only set to `true` if overviews are used. If it is `false` or it's `undefined` then `stats.usesOverviews` is not set.

In other words: We log only `stats.usesOverviews` if it is `true`

enekid:
The part of logging the map instantiations that use overviews is still missing. WIP.

lib/cartodb/models/dataview/base.js
dgaubert: Ok, got it. @enekid explain it to me off-line

oleurud: It will be easy to use Kibana if you log the `false` too

Issue #1037, opened 2018-09-17
Aggregation dimension classifications


jgoizueta:
We need to add `dimensions` control to the aggregation API, so that, in addition to the spatial aggregation (controlled by `resolution` and `placement`) we can add additional expressions into the GROUP BY of the aggregation.

We don't want to allow arbitrary SQL expressions for the dimensions. We'd like to support categorical columns and classified continues values. For the latter, they could be numbers or times and we need to control the granularity of the classification, e.g. if the time is shown as hours, days, months... or if prices are shown in 3 intervals, etc.

We'll start by implementing the time case and validate the API by [integrating its use in CARTO VL](https://github.com/CartoDB/carto-vl/issues/551)

For time aggregation, there are two interesting cases:
* Grouping a time series into time units such as minutes, hours, days, months, years. A refinement of this would be two allow groups of N units, e.g. 10-minute buckets, 4-month, etc.
* Grouping as before into time units, but computing them relative to (modulo) a larger time unit: hours of the day (0-23), days of the week (0-6), months of the year (1-12), ...

### API parameters

We could give support to these needs with an API like this: (`dimensions` would be a parameter within the `aggregation` of the layer)

```javascript
dimensions: {
  hour_of_day: {
    column: 'time_of_arrival'
    granularity: 'hour',
    multiplicity: 1, // optional: number of units per grain
    cycle: 'day',    // optional: granularity relative to this cycle
    offset: 0,       // optional: starting value for granularity
    timezone: -7200  // optional: UTC offset in seconds of the result
  }
}
```

Note: this nomenclature is unfortunate, because the date histogram already uses `offset` for the time zone offset. So we must review this. (if for example we decide not to support the offset defined here, see below, be could use  `offset` for the timezone offset for consistency with histograms).

The offset parameter could be useful to specify both the order of cyclic units and how they are grouped:
* With multiplicity > 1, e.g. 3 for months/year, by default we would get quarters as jan-feb-mar, may-apr-jun, etc.
  If we use an offset of 1, we would get feb-mar-may, ... nov-dec-jan instead.
* With multiplicity = 1 it would simply define the starting point of the cycle, e.g. what day of the week is the first (0), sunday, monday?

The timezone parameter is necessary because it determines how time is broken up into days, etc. The user would specify in which timezone is the classification to be performed.

The general SQL expression for the classification would be equivalent to:
[...]

PR #1039, opened 2018-10-03
Time dimensions

jgoizueta:
See #1037 (WIP)

### DIFF
--- a/docs/aggregation.md
+++ b/docs/aggregation.md
@@ -140,7 +140,7 @@ By default aggregated data is grouped only by the the spatial discretization int
 The `dimensions` parameters permits to define additional aggregation dimensions. Currently only date (time, timestamp) columns can be used as dimensions. Just like the spatial position is discretized using a grid, for every
 additional dimension, a grouping into discrete periods of time must be defined.
 
-```json
+```js
 {
     "dimensions": {
         "time_in_days": { // Name of the resulting dimension column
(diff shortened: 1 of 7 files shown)


jgoizueta:
Hey @dgaubert  it would be nice that someone starts to review this, the CARTO VL front to expose this ain't finished yet but I don't forsee relevant changes here. There's some documentation [here](https://github.com/CartoDB/Windshaft-cartodb/blob/a4dfc09c719b3f78e08ed20d62c93f200a3394e4/docs/aggregation.md#dimensions):

docs/aggregation.md
```diff
@@ -134,6 +134,61 @@ of the original dataset applying three different aggregate functions.
...
+### `dimensions`
+
+By default aggregated data is grouped only by the the spatial discretization into grid cells.
+The `dimensions` parameters permits to define additional aggregation dimensions. Currently only date (time, timestamp) columns can be used as dimensions. Just like the spatial position is discretized using a grid, for every
+additional dimension, a grouping into discrete periods of time must be defined.
+
+```json
+{
+    "dimensions": {
+        "time_in_days": { // Name of the resulting dimension column
```
jgoizueta: 🤔 I should either remove the comments or change it to javascript

dgaubert: use `js` instead.

lib/cartodb/backends/layer-stats/mapnik-layer-stats.js
```diff
@@ -178,6 +201,18 @@ function _columnStats(ctx, columns) {
                         stats[name][fn] = res.rows[0][`${name}_${fn}`];
                     });
                 });
+                Object.keys(dimensionsInfo).forEach(name => {
+                    // Temporalily place dimensions info in stats.columns.__dimensions
+                    stats.__dimensions = stats.__dimensions || {};
```
dgaubert: Why do we need `__dimensions` as temporally placeholder?. Are we hacking a bit `stats` to prepare dimensions statistics later?

lib/cartodb/backends/layer-stats/mapnik-layer-stats.js
```diff
@@ -137,35 +142,53 @@ function _sample(ctx, numRows) {
     return Promise.resolve();
 }
 
-function _columnStats(ctx, columns) {
+function _columnStats(ctx, columns, dimensions) {
     if (!columns) {
         return Promise.resolve();
     }
-    if (ctx.metaOptions.columnStats) {
+    if (ctx.metaOptions.columnStats || ctx.metaOptions.dimensions) {
```
dgaubert: I'd love to see some refactor here, it's getting hard to understand what's happening. I think by extracting some common functionality and then splitting conditionals would be enough.

lib/cartodb/backends/layer-stats/mapnik-layer-stats.js
jgoizueta: I got into this ugly mess ad I couldn't find a better way out: Here we initially collected info about columns only and all the results are gathered at the end of this function and merged into a `columns` property:
```
return Promise.all(queries).then(results => ({ columns: mergeColumns(results) }));
```
But now I also need to put some of the results in a `dimensions` property, so I use `__dimensions` in hope no column will be ever named like that 🤞 and later, where all the queries are executed it is moved to where it should be.

I'll give it another thought with a fresher mind tomorrow 😅

PR #1041, opened 2018-10-09
Sample configs: use PostGIS to generate MVT's

rafatower:
This is all about testing of `pg-mvt` renderer. The changes are explained in the commit messages, but to sum up:
- The docker image is updated to use latest postgis v2.4.4.5-carto.1
- The config examples now have `mvt.usePostGIS` set to `true`. They are used to generate the config for testing and CI.
- Some tests have been slightly modified. Namely:
  - the `pg-mvt` renderer does not generate feature `id`'s. That is something added by `toGeoJSONSync` (that is, an artifact from the tests). I took into consideration the difference in behavior with respect to mapnik itself, and asked carto-vl guys and they are ok with that behavior. BTW, that behavior is loosely specified in the MVT specs.
  - the scale denominator hardcoded in one test is tweaked. Postgis simply calculates it with more precision.
  - one test case used a point in the World's border. Mapnik did not render it. That is arguably a bug in mapnik.


docker/Dockerfile-nodejs6-xenial-pg101:postgis-2.4
```diff
@@ -0,0 +1,89 @@
+FROM ubuntu:xenial
```
dgaubert: Is this file being used? Or we want to keep it for some reason...

oleurud: Of course, I don't like this versioning mode (create a docker file for every tag). I only used this way to move forward quickly

rafatower: todos en "violento acuerdo", moving forward :smiley:

PR #1046, opened 2018-10-19
Do not use 'Object.assign' as '_.defautls' equivalent

dgaubert:


oleurud:
What is the reason for this? I think the only difference between both cases is using `Object.assing` the `tplValues` could have more properties. For example:

```javascript
params = {a:1, b:2, c:3}
// assign
assign_params = Object.assign({a:11, b:12}, params) 
// { a: 1, b: 2, c: 3 }
// If we could only need `a` and `b`, but we will have `c` too
```

Is it a problem?

Anyway, the new code also looks nice to me.

dgaubert:
`Object.assign` doesn't work as we were expecting: 
```
$ node
> const obj = Object.assign({ a: '', b: '' }, { a: 1, b: undefined });
undefined
> obj
{ a: 1, b: undefined }
> const { a = '', b = '' } = { a: 1, b: undefined };
undefined
> a
1
> b
''
>
```

oleurud:
Ok, good to know

dgaubert:
Now, it's compatible with `_.defaults`:

```
$ node
> const obj = Object.assign({ a: '', b: '' }, { a: 1, b: undefined });
undefined
> obj
{ a: 1, b: undefined }
> const { a = '', b = '' } = { a: 1, b: undefined };
undefined
> a
1
> b
''
> const _ = require('underscore');
undefined
> const __obj = _.defaults({ a: 1, b: undefined }, {a: '', b: '' });
undefined
> __obj
{ a: 1, b: '' }
>
```

PR #1050, opened 2018-10-23
Prevent "Cannot read property 'geom_type' of undefined" on layer stats

rochoa:
The server responds with a 400 error due to a TypeError, this avoids that and returns a valid response without the geometryType metadata.

This will help with https://github.com/CartoDB/carto-vl/issues/1049.


jgoizueta:
I don't quite understand it, but maybe we need use the `serverOptions` helper like here: https://github.com/CartoDB/Windshaft-cartodb/blob/50e232ae4020f35dc6c47c2b4e402e0e94f67b1b/test/acceptance/aggregation.js#L290-L291

PR #1052, opened 2018-10-23
Use strict mode

dgaubert:
Use strict mode for all modules


dgaubert:
I've just taken care of it. Should pass now...

PR #1070, opened 2019-02-12
Dynamic node version in Dockerfile 

oleurud:
- Create a new docker image allowing us to work with any Node.js version
- Use the image


Algunenano:
I know it's out of scope, but it'd be really nice if this was also prepared to choose Postgresql version (10 and 11), as otherwise we'll have to repeat it in the next cycle.

oleurud:
We will update the Node.js version often and it becomes in an easy CI task using `nvm`. `nvm` allows us to select the Node.js version when we call the `docker run` task. 

In the Postgres side, I don't see an easy way to create a dynamic selector of the PG version, but it is not a big issue to create a new docker version for each version.

.travis.yml
```diff
@@ -4,14 +4,14 @@ jobs:
       services:
         - docker
       language: generic
-      before_install: docker pull carto/nodejs6-xenial-pg101:postgis-2.4.4.5
-      script: npm run docker-test -- nodejs6
+      before_install: docker pull carto/nodejs-xenial-pg101:latest
+      script: npm run docker-test -- 6
```
dgaubert:
```suggestion
      script: npm run docker-test -- 6.9.2
```

run_tests_docker.sh
```diff
@@ -1,40 +1,25 @@
...
-NODEJS_VERSION=${1-nodejs10}
-
-if [ "$NODEJS_VERSION" = "nodejs10" ];
-then
-    echo "npm version on install:"
-    npm -v
-    npm ci
-    npm ls
-elif [ "$NODEJS_VERSION" = "nodejs6" ];
+if [ "$NODEJS_VERSION" = "6" ];
```
dgaubert:
```suggestion
if [ "$NODEJS_VERSION" = "6.9.2" ];
```

PR #1072, opened 2019-02-13
Upgrade camshaft 0.63.4

rafatower:
This fixes https://github.com/CartoDB/camshaft/issues/380

Will test in staging, merge and release as soon as it's tested (with the procedure in the ticket)


rafatower:
Tested in staging. Now it works as expected

image

(or more exactly: it fails in a controlled way, with a proper error message, instead of raising an `Uncaught exception`)

PR #1076, opened 2019-02-21
Drop suppor for Node.js 6, npm 3, yarn and redis 3

dgaubert:
- Drop support for Node.js 6
- Drop support for npm 3
- Stop supporting `yarn.lock`
- Drop support for Postgres 9.5
- Drop support for PosGIS 2.2
- Drop support for Redis 3


INSTALL.md
```diff
@@ -4,32 +4,21 @@
...
-  - Redis >2.4.0 (http://www.redis.io)
-  - Mapnik >3.x. See [Installing Mapnik](https://github.com/CartoDB/Windshaft#installing-mapnik).
+  - Node 10.x
+  - npm 6.x
+  - gcc 4.9
+  - PostgreSQL >= 10.0
+  - PostGIS >= 2.4
+  - CartoDB Postgres Extension >= 0.24.1
+  - Redis >= 4
+  - Mapnik == 3.0.15.9. See [Installing Mapnik](https://github.com/CartoDB/Windshaft#installing-mapnik).
```
Algunenano: Do we need to include Mapnik here considering that we expect it to be installed as part of `npm ci`? You don't need it installed in your system as a library.

dgaubert: Right, I was keeping it for historical reasons. Removing it

dgaubert: done

Algunenano:
The PARALLEL workaround in test/support/prepare_db.sh should be removed too. Also any comments referencing Postgis before 2.4 (use_postgis, features, etc).

I haven't checked yet if there are other things that depend on those versions.

dgaubert:
@Algunenano, are we ready?

Algunenano:
The only thing I can see missing is the update of the environment samples (all 4):

This comment can go away:
```
      mvt: {
        //If enabled, MVTs will be generated with PostGIS directly, instead of using Mapnik,
        //PostGIS 2.4 is required for this to work
        //If disabled it will use Mapnik MVT generation
        usePostGIS: true
      },
```

Same as this (should it be on now by default?):
```
          // SQL queries will be wrapped with ST_ClipByBox2D
          // Returning the portion of a geometry falling within a rectangle
          // It will only work if snapToGrid is enabled
          clipByBox2d: false, // this requires postgis >=2.2 and geos >=3.5
```

PR #1077, opened 2019-03-01
Experimental support for listing features in a grid

dgaubert:


dgaubert:
TODO: 
 -  Reduce complexity in method `ClusterBackend#getClusterFeatures`
 - Set a rate limit for the new endpoint
 - Configure cache layers for the new endpoint, ¿?

Issue #1074, opened 2019-02-18
Histogram performance improvement


Algunenano:
In the histogram SQL query, the `__cdb_iqrange` table is declared as a CTE but only used once. A quick  test show a query time reduced from 1.3 s to 1.0 when inlining it instead (obviously this depends on the data)

cc/ @CartoDB/rt-managers

javitonino:
Removing this from RT. This is not urgent. If you feel this is important, let's do a project for it.

PR #1080, opened 2019-03-04
Numeric histogram performance improvement

Algunenano:
Up to 2.5x as fast as the current implementation.
- Inlines `__cdb_filtered_source`.
- Uses `percentile_disc` instead of `ntile` to calculate the interquantile range.

Some benchmarks (the response is exactly the same for all 6):
[hist_master_vs_hist_improved.pdf](https://github.com/CartoDB/Windshaft-cartodb/files/2926994/hist_master_vs_hist_improved.pdf)

Closes https://github.com/CartoDB/Windshaft-cartodb/issues/1074

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,7 @@ Released 2019-mm-dd
 
 Announcements:
 - Experimental support for listing features in a grid when the map uses the dynamic agregation.
+- Numeric histogram performance improvement (#1080)
 - Update deps:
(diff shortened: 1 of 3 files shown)


dgaubert:
@jgoizueta, can you help us to move this PR forward?

Algunenano:
> It's nice how percentile_disc simplifies the query and makes it faster. I'd only wish that PG took care of optimizing away CTEs in cases like this rather than forcing us to do so.

It will be done in PG12.

PR #1082, opened 2019-03-06
Fix boolean aggregation layer option

### LINKED ISSUE #1073: Aggregation:false doesn't work
I'm testing a big dataset (> aggregation threshold, 683788 points vs 100k) with `aggregation:false` and it's still aggregating the layer.

Full mapConfig:
```
{"layergroup": {"buffersize": {"mvt": 1}, "layers": [{"options": {"sql":"SELECT * FROM benchmark_eb6e5663e0cac56038cae186f5c59169ec828bf02f7340895c1d8", "vector_extent":2048, "aggregation":false}, "type": "mapnik"}],"version": "1.0.1"},"name": "tpl_d8411740_311d_11e9_a060_9c5c8ed39a6e","version": "0.0.1"}
```

I've found that:
a) This doesn't happen if the map is smaller than the default threshold.
b) This doesn't happen if I pass a string instead of a boolean, that is instead of passing `"aggregation":false` I pass `"aggregation":"false"`.

I've tried changing https://github.com/CartoDB/Windshaft-cartodb/blob/e93dd982b9e4ba46d69fbe9ec528b997624121d5/lib/cartodb/models/mapconfig/adapter/aggregation-mapconfig-adapter.js#L52 to include `aggregation === false` without any luck.

cc/ @CartoDB/rt-managers

jgoizueta:
Fixes #1073 

Before this fix, vector-only MapConfig layers where always aggregated (except for number of rows threshold or type of geometry reasons) even if `aggregated` was set to `false`.

I've found handling of aggregation enablement slightly confusing in the code. Some notes about how things are after this fix:
* `AggregationMapConfig` deals with the requirement to aggregate layers:
  + `hasLayerAggregation(i)` says if a layer has explicitly defined aggregation (and now returns a special value if they're explicitly disabled).
  + `isAggregationLayer(i)` determines if the layer requires aggregation, taking also into account if the MapConfig is vector-only.
  + `getAggregation(i)` returns falsy objects (undefined, false) only if no aggregation should be performed. Otherwise it returns {} if aggregation wasn't explicitly defined or was true.
* `AggregationMapConfigAdapter` (`_shouldAdaptLayer`) uses this information and takes into account the threshold/table size and type of geometry to decide if aggregations are finally performed.

Do not be confused (as I was) with the optional request parameter `aggregation`. In `AggregationMapConfigAdapter`, `_isValidAggregationQueryParam` and `_shouldAdapt` look at this parameter (not the layer options one) which, since it comes from the request, it's a string ('true'/'false') and is used to prevent (or force) aggregation handling for the whole MapConfig.

Passing a string (e.g. 'false' or 'true') in the layer options `aggregation` is not supported (anything other than a object or boolean is considered as `undefined`). We should probably validate that parameter and produce and error for unsupported types.

@dgaubert I leave this here as your welcome-back-to-work present


lib/cartodb/models/aggregation/aggregation-mapconfig.js
```diff
@@ -49,6 +49,10 @@ module.exports = class AggregationMapConfig extends MapConfig {
         ];
     }
 
+    static get HAS_AGGREGATION_DISABLED () {
```
dgaubert: I don't understand well this static property, I guess you are trying to improve the readability to determine whether we have aggregation or not. However, the name is not meaningful:

`hasAggregation !== AggregationMapConfig.HAS_AGGREGATION_DISABLED`

vs

`hasAggregation !== null`

In any case, I don't have a strong opinion.

lib/cartodb/models/aggregation/aggregation-mapconfig.js
```diff
@@ -103,42 +107,50 @@ module.exports = class AggregationMapConfig extends MapConfig {
...
+     * * AggregationMapConfig.HAS_AGGREGATION_DISABLED: explicitly disabled aggregation (false)
+     */
     hasLayerAggregation (index) {
         const layer = this.getLayer(index);
         const { aggregation } = layer.options;
-
-        return aggregation !== undefined && (typeof aggregation === 'object' || typeof aggregation === 'boolean');
+        if (aggregation !== undefined && (typeof aggregation === 'object' || typeof aggregation === 'boolean')) {
+            if (aggregation === false) {
+                return AggregationMapConfig.HAS_AGGREGATION_DISABLED;
```
dgaubert: shouldn't we return a `false` value?. Why do we need to return a `null` here?

lib/cartodb/models/aggregation/aggregation-mapconfig.js
```diff
@@ -103,42 +107,50 @@ module.exports = class AggregationMapConfig extends MapConfig {
...
     }
 
     getAggregation (index) {
-        if (this.isVectorOnlyMapConfig() && !this.hasLayerAggregation(index)) {
-            return {};
-        }
-
         const { aggregation } = this.getLayer(index).options;
 
-        if (typeof aggregation === 'boolean') {
```
dgaubert: 😱

dgaubert: Thanks for fixing this

test/support/sql/windshaft.test.sql
```diff
@@ -819,5 +819,46 @@ CREATE INDEX test_table_100_the_geom_webmercator_idx ON test_table_100 USING gis
 GRANT ALL ON TABLE test_table_100 TO :TESTUSER;
 GRANT SELECT ON TABLE test_table_100 TO :PUBLICUSER;
 
+-- Table with 200K rows
+CREATE TABLE test_table_200k (
```
dgaubert: ❤️

test/acceptance/aggregation.js
```diff
@@ -690,6 +694,66 @@ describe('aggregation', function () {
                 });
             });
 
+            it('use default aggregation by setting `aggregation: true`', function (done) {
```
dgaubert: What about adding a test to validate the contrary `aggregation: false`? 

Even I'd add a test to validate `aggregation: 'false'` should not work

lib/cartodb/models/aggregation/aggregation-mapconfig.js
jgoizueta: Now `hasLayerAggregation` has three possible values:
* true: explicit aggregation, can be `aggregation: {}` or `aggregation: true`.
* false:  explicit aggregation, i.e. undefined
* SOMETHING_ELSE: explicitly disabled `aggregation: false`.

So to distinguish the new third option I needed some value, I chose `null`, but I wanted it to be explicit and use a constant to make it clear, and ended up using that static property 🤷‍♂️ (and the name is `HAS_AGGREGATION_DISABLED`)

test/acceptance/aggregation.js
jgoizueta: We already had one for aggregation: false but was ineffective because the query used had fewer rows than the threshold, so it was not aggregated, but not because of the false. Now it's fixed with the `test_table_200k` table.

For the string case I think we should not only add tests, but change the behaviour: now strings (anything not an object or boolean) are just ignored, as if aggregation was not present, and I think it would be better to return a validation error. What do you think?

dgaubert: Fair enough! :)

Algunenano:
Can this be merged?

jgoizueta:
I vote yes. But @dgaubert has the last word. 

Note: regarding the tests for `aggregation: 'false'` my position is that since the right thing to do is tho change behaviour and reject strings in that parameter it should be handled in another ticket.

dgaubert:
You have my blessings!

PR #1089, opened 2019-03-26
Implement graceful shutdown

dgaubert:
Related https://github.com/CartoDB/cartodb-platform/issues/5389

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,14 +4,19 @@
 Released 2019-mm-dd
 
 Announcements:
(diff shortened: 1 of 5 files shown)


dgaubert:
> I see you have not added the SIGHUP, but AFAIK, we are not going to send it, so 👍

SIGHUP was previously handled [here](https://github.com/CartoDB/Windshaft-cartodb/blob/master/app.js#L172-L178)

PR #1107, opened 2019-07-03
Set directive 'max-age' to a fallback value

dgaubert:
Set directive 'max-age' to a fallback value when there are affected tables where we can't know when were updated for the last time, e.g: non-cartodified tables or foreign tables without cartodb support

### DIFF
--- a/lib/cartodb/api/middlewares/cache-control-header.js
+++ b/lib/cartodb/api/middlewares/cache-control-header.js
@@ -2,7 +2,7 @@
 
 const ONE_YEAR_IN_SECONDS = 60 * 60 * 24 * 365;
 const FIVE_MINUTES_IN_SECONDS = 60 * 5;
-const FALLBACK_TTL = global.environment.varnish.fallbackTtl || FIVE_MINUTES_IN_SECONDS
+const FALLBACK_TTL = global.environment.varnish.fallbackTtl || FIVE_MINUTES_IN_SECONDS;
 
 module.exports = function setCacheControlHeader ({
     ttl = ONE_YEAR_IN_SECONDS,
@@ -22,7 +22,7 @@ module.exports = function setCacheControlHeader ({
                 return next();
             }
 
-            const directives = [ 'public' ]
+            const directives = [ 'public' ];
 
             if (affectedTables && !affectedTables.getTables().some(table => !!table.updated_at)) {
                 directives.push(`max-age=${fallbackTtl}`);
--- a/lib/cartodb/api/middlewares/cache-control-header.js
+++ b/lib/cartodb/api/middlewares/cache-control-header.js
@@ -24,7 +24,7 @@ module.exports = function setCacheControlHeader ({
 
             const directives = [ 'public' ];
 
-            if (affectedTables && !affectedTables.getTables().some(table => !!table.updated_at)) {
+            if (everyAffectedTablesCanBeInvalidated(affectedTables)) {
                 directives.push(`max-age=${fallbackTtl}`);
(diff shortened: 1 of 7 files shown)


Algunenano:
Could you add some test? I understand that this behaviour should affect all non-cartodbfied tables (local or remote), so I'd test with a map using one of those tables and a map using a JOIN between different tables (with and without information in cdb_tablemetadata).

test/acceptance/cache/cache-control-header.js
```diff
@@ -0,0 +1,91 @@
...
+            testClient.getTile(0, 0, 0, {}, function (err, res) {
+                if (err) {
+                    return done(err);
+                }
+
+                assert.equal(res.headers['cache-control'], `public,max-age=${ttl}`);
+                testClient.drain(done);
+            });
+        });
+
```
Algunenano: It would be nice to also test a query that uses multiple tables (some in cdb_tablemetada and some that aren't) to check that in that case it's also 5 minutes.

dgaubert: I'm working on it. :billy_the_kid:

😄

test/acceptance/cache/cache-control-header.js
```diff
@@ -0,0 +1,124 @@
...
+function createMapConfig (layers = defaultLayers) {
+    return {
+        version: '1.8.0',
+        layers: layers
+    };
+}
+
+describe('cache-control header', function () {
+    describe('max-age directive', function () {
+        it('tile from a table wich is included in cdb_tablemetada', function (done) {
```
Algunenano: wich/which

Algunenano: Same in the next text

dgaubert: 🤦‍♂

lib/cartodb/api/middlewares/cache-control-header.js
```diff
@@ -1,21 +1,46 @@
...
-        }
+        mapConfigProvider.getAffectedTables((err, affectedTables) => {
+            if (err) {
+                global.logger.warn('ERROR generating Cache Control Header:', err);
+                return next();
+            }
+
+            const directives = [ 'public' ];
+
+            if (everyAffectedTablesCanBeInvalidated(affectedTables)) {
```
Algunenano: Is this name correct? I'm reading "If all tables in the query can be invalidated then I use 5 minutes, if not I use 1 year" which is the opposite of what is doing.

dgaubert: what a long week... thanks! you are totally right!

lib/cartodb/api/middlewares/cache-control-header.js
```diff
@@ -42,5 +42,5 @@ module.exports = function setCacheControlHeader ({
 };
 
 function everyAffectedTablesCanBeInvalidated (affectedTables) {
-    return affectedTables && affectedTables.getTables().some(table => !table.updated_at);
+    return affectedTables && affectedTables.getTables().every(table => !!table.updated_at);
```
Algunenano: Being picky: `!!table.updated_at` will also reject 0, even though it should be a valid value.

lib/cartodb/api/middlewares/cache-control-header.js
rafatower: Just as a note for the future: "can be invalidated" depends on the one triggering the invalidation, and we're inferring it from the info or absence of `updated_at`

PR #1111, opened 2019-07-10
Render MVTs and aggregations without cartodb-postgresql

Algunenano:
- [WIP] Updates Windshaft to current master.
- Overviews adapter: Do not check for overviews if the map is MVT only or has aggregations enabled.
- Use windshaftUtils.WebMercatorHelper() to calculate tile boundaries.
- Use `!scale_denominator! * 0.00028` to calculate tile resolution. This works both with pg-mvt and Mapnik (where other variables depend on tile size, typically 2048).
- Updates Mapnik as part of Windshaft where the precision of variables has been improved.
- Use native Postgres mode() instead of _cdb_mode.
- Changes how aggregations are calculated to improve performance. Note that now there is only a single query template.
- A couple of test have been changed to not check for exact tile output, and instead check specific conditions (like non-partially aggregated cells).

### DIFF
--- a/lib/cartodb/backends/cluster.js
+++ b/lib/cartodb/backends/cluster.js
@@ -157,8 +157,7 @@ const clusterFeaturesQuery = ctx => `
 
 const gridResolution = ctx => {
     const minimumResolution = 2*Math.PI*6378137/Math.pow(2,38);
-    const pixelSize = `CDB_XYZ_Resolution(${ctx.zoom})`;
-    return `GREATEST(${256/ctx.res}*${pixelSize}, ${minimumResolution})::double precision`;
+    return `${256/ctx.res} * GREATEST(!pixel_height!, ${minimumResolution})::numeric`;
 };
 
 const aggregationQuery = ctx => `
--- a/lib/cartodb/backends/cluster.js
+++ b/lib/cartodb/backends/cluster.js
@@ -127,7 +127,7 @@ function getClusterFeatures (pg, zoom, clusterId, columns, query, resolution, ag
     } , true); // use read-only transaction
 }
 
-const schemaQuery = ctx => `SELECT * FROM (${ctx.query}) __cdb_schema LIMIT 0`;
+const schemaQuery = ctx => `SELECT * FROM (${ctx.query}) __cdb_cluster_schema LIMIT 0`;
 const clusterFeaturesQuery = ctx => `
     WITH
     _cdb_params AS (
--- a/lib/cartodb/backends/cluster.js
+++ b/lib/cartodb/backends/cluster.js
@@ -5,6 +5,9 @@ const dbParamsFromReqParams = require('../utils/database-params');
 const debug = require('debug')('backend:cluster');
 const AggregationMapConfig = require('../models/aggregation/aggregation-mapconfig');
 
+const windshaftUtils = require('windshaft').utils;
+const wmh = new windshaftUtils.WebMercatorHelper();
+
 module.exports = class ClusterBackend {
     getClusterFeatures (mapConfigProvider, params, callback) {
         mapConfigProvider.getMapConfig((err, _mapConfig) => {
@@ -156,8 +159,8 @@ const clusterFeaturesQuery = ctx => `
 `;
 
 const gridResolution = ctx => {
-    const minimumResolution = 2*Math.PI*6378137/Math.pow(2,38);
-    return `${256/ctx.res} * GREATEST(!pixel_height!, ${minimumResolution})::numeric`;
+    const zoom_resolution = wmh.getResolution({ z : Math.min(38, ctx.zoom) });
+    return `${256/ctx.res} * (${zoom_resolution})::double precision`;
 };
 
 const aggregationQuery = ctx => `
--- a/lib/cartodb/backends/cluster.js
+++ b/lib/cartodb/backends/cluster.js
@@ -6,7 +6,7 @@ const debug = require('debug')('backend:cluster');
 const AggregationMapConfig = require('../models/aggregation/aggregation-mapconfig');
 
 const windshaftUtils = require('windshaft').utils;
-const wmh = new windshaftUtils.WebMercatorHelper();
+const webmercator = new windshaftUtils.WebMercatorHelper();
 
 module.exports = class ClusterBackend {
     getClusterFeatures (mapConfigProvider, params, callback) {
@@ -159,8 +159,8 @@ const clusterFeaturesQuery = ctx => `
 `;
 
 const gridResolution = ctx => {
-    const zoom_resolution = wmh.getResolution({ z : Math.min(38, ctx.zoom) });
-    return `${256/ctx.res} * (${zoom_resolution})::double precision`;
+    const zoomResolution = webmercator.getResolution({ z : Math.min(38, ctx.zoom) });
+    return `${256/ctx.res} * (${zoomResolution})::double precision`;
 };
 
 const aggregationQuery = ctx => `
--- a/lib/cartodb/models/aggregation/aggregation-query.js
+++ b/lib/cartodb/models/aggregation/aggregation-query.js
@@ -310,26 +310,39 @@ const gridResolution = ctx => {
 };
 
 // SQL query to extra the boundaries of the area to be aggregated and the grid resolution
+// cdb_{x-y}{min_max} return the limits of the tile. Aggregations do [min, max) in both axis
+// cdb_res: Aggregation resolution (as specified by gridResolution)
+// cdb_point_bbox: Tile bounding box [min, max]
 const gridInfoQuery = ctx => {
     return `
     SELECT
-        res,
-        CEIL (ST_XMIN(cdb_full_bbox) / res) * res AS cdb_xmin,
-        FLOOR(ST_XMAX(cdb_full_bbox) / res) * res AS cdb_xmax,
-        CEIL (ST_YMIN(cdb_full_bbox) / res) * res AS cdb_ymin,
-        FLOOR(ST_YMAX(cdb_full_bbox) / res) * res AS cdb_ymax
+        cdb_xmin,
+        cdb_ymin,
+        cdb_xmax,
+        cdb_ymax,
+        cdb_res,
+        ST_MakeEnvelope(cdb_xmin, cdb_ymin, cdb_xmax, cdb_ymax, 3857) AS cdb_point_bbox
     FROM
     (
         SELECT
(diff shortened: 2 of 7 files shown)


lib/cartodb/models/aggregation/aggregation-query.js
```diff
@@ -297,152 +289,158 @@ const havingClause = ctx => {
...
+        ) __cdb_src_get_x_y, __cdb_grid_params
+        WHERE cdb_x < __cdb_grid_params.cdb_xmax AND cdb_y < __cdb_grid_params.cdb_ymax
+    ) __cdb_src_gridded
+    GROUP BY cdb_pos_grid_x, cdb_pos_grid_y ${dimensionNames(ctx)}
+    ${havingClause(ctx)}
+) __cdb_aggregation_src
+${aggregatedPointJoin(ctx)}
 `;
 
 const aggregationQueryTemplates = {
```
rafatower: How about removing this? I think now you can replace its usages by `defaultAggregationQueryTemplate` and define `SUPPORTED_PLACEMENTS` explicitly.

OTOH I personally find the separate templates slightly more readable, but no strong opinion :man_shrugging:

lib/cartodb/models/aggregation/aggregation-query.js
```diff
@@ -297,152 +289,158 @@ const havingClause = ctx => {
...
+            return ``;
+
+        case `point-grid`:
+            return ``;
+
+        // For point-sample we'll get a single point directly from the source
+        // If it's default aggregation we'll add the extra columns to keep backwards compatibility
+        case `point-sample`:
+            return `NATURAL JOIN
+(
```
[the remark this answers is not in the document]
dgaubert: IMO we should pick the indentation which makes easier to read in code rather than logs.

test/acceptance/aggregation.js
```diff
@@ -6,6 +6,9 @@ const assert = require('../support/assert');
 const TestClient = require('../support/test-client');
 const serverOptions = require('../../lib/cartodb/server_options');
 
+const windshaftUtils = require('windshaft').utils;
+const wmh = new windshaftUtils.WebMercatorHelper();
```
dgaubert: I'd try to avoid acronyms as variable names.

lib/cartodb/backends/cluster.js
```diff
@@ -156,9 +159,8 @@ const clusterFeaturesQuery = ctx => `
 `;
 
 const gridResolution = ctx => {
-    const minimumResolution = 2*Math.PI*6378137/Math.pow(2,38);
-    const pixelSize = `CDB_XYZ_Resolution(${ctx.zoom})`;
-    return `GREATEST(${256/ctx.res}*${pixelSize}, ${minimumResolution})::double precision`;
+    const zoom_resolution = wmh.getResolution({ z : Math.min(38, ctx.zoom) });
```
dgaubert: Camel Case

lib/cartodb/models/aggregation/aggregation-query.js
```diff
@@ -297,152 +289,158 @@ const havingClause = ctx => {
...
 // the pixel size of zoom level 30 (i.e. 1/2*(30+8) of the full earth web-mercator extent), which is about 0.15 mm.
-// Computing this using !scale_denominator!, !pixel_width! or !pixel_height! produces
-// inaccurate results due to rounding present in those values.
+//
+// NOTE: We'd rather use !scale_denominator!, but in Mapnik this value is extent / 256 for raster
+// and extent / tile_extent {4096 default} for MVT, so since aggregations are always based
+// on 256 we can't have the same query in both cases
+// As this scale change doesn't happen in !scale_denominator! we use that instead
 const gridResolution = ctx => {
     const minimumResolution = 2*Math.PI*6378137/Math.pow(2,38);
```
jgoizueta: We could use WebMercatorHelper here too for minimum resolution

lib/cartodb/models/aggregation/aggregation-query.js
```diff
@@ -297,152 +289,158 @@ const havingClause = ctx => {
 // (i.e. each tile is divided into ctx.res*ctx.res cells).
 // We limit the the minimum resolution to avoid division by zero problems. The limit used is
 // the pixel size of zoom level 30 (i.e. 1/2*(30+8) of the full earth web-mercator extent), which is about 0.15 mm.
-// Computing this using !scale_denominator!, !pixel_width! or !pixel_height! produces
-// inaccurate results due to rounding present in those values.
+//
+// NOTE: We'd rather use !scale_denominator!, but in Mapnik this value is extent / 256 for raster
```
jgoizueta: Do you mean `!pixel_width!` here?

lib/cartodb/models/aggregation/aggregation-query.js
```diff
@@ -297,152 +289,158 @@ const havingClause = ctx => {
...
+// and extent / tile_extent {4096 default} for MVT, so since aggregations are always based
+// on 256 we can't have the same query in both cases
+// As this scale change doesn't happen in !scale_denominator! we use that instead
 const gridResolution = ctx => {
     const minimumResolution = 2*Math.PI*6378137/Math.pow(2,38);
-    const pixelSize = 'CDB_XYZ_Resolution(CDB_ZoomFromScale(!scale_denominator!))';
-    return `GREATEST(${256/ctx.res}*${pixelSize}, ${minimumResolution})::double precision`;
+    return `${256/ctx.res} * GREATEST(!scale_denominator! * 0.00028, ${minimumResolution})::double precision`;
```
jgoizueta: I would include a comment to clarify where the magic number 0.00028 comes from... something like:
> 0.00028 is the conventional pixel size in meters used to compute `!scale_denominator!`

By the way, I wonder where does it really come from... I see that value mentioned [here](https://en.wikipedia.org/wiki/Dot_pitch) as a "typical entry-level monitor dot pitch", but it refers to the diagonal, not the width or height of the pixel... so maybe it all comes from a misunderstanding ¯\_(ツ)_/¯

lib/cartodb/models/aggregation/aggregation-query.js
```diff
@@ -297,152 +289,158 @@ const havingClause = ctx => {
...
-//   Xmin <= x <= Xmax and Ymin <= y <= Ymax
-// So, in order to be both efficient and accurate we will need to use both
-// conditions for spatial filtering.
-const spatialFilter = `
-  (_cdb_query.the_geom_webmercator && _cdb_params.bbox) AND
-    ST_X(_cdb_query.the_geom_webmercator) >= _cdb_params.xmin AND
-    ST_X(_cdb_query.the_geom_webmercator) < _cdb_params.xmax AND
-    ST_Y(_cdb_query.the_geom_webmercator) >= _cdb_params.ymin AND
-    ST_Y(_cdb_query.the_geom_webmercator) < _cdb_params.ymax
+// SQL query to extra the boundaries of the area to be aggregated and the grid resolution
```
jgoizueta: typo extra -> extract

lib/cartodb/models/aggregation/aggregation-query.js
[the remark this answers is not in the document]
Algunenano: The logs (or at least mine) don't show indentation, so this is me making it easier to read. I'm open to suggestions as to how to reformat to improve readability.

test/acceptance/aggregation.js
```diff
@@ -96,9 +99,9 @@ describe('aggregation', function () {
     const POLYGONS_SQL_1 = `
     select
         x + 4 as cartodb_id,
-        st_buffer(st_setsrid(st_makepoint(x*10, x*10), 4326)::geography, 100000)::geometry as the_geom,
+        st_buffer(st_setsrid(st_makepoint(x*10, x*10), 4326), 10) as the_geom,
```
jgoizueta: Avoiding the geography type for any particular reason?

test/acceptance/aggregation.js
```diff
@@ -3125,48 +3104,41 @@ describe('aggregation', function () {
...
+                                        { cartodb_id: 3, _cdb_feature_count: 1 },
+                                        { cartodb_id: 4, _cdb_feature_count: 2 },
+                                        { cartodb_id: 5, _cdb_feature_count: 2 },
+                                        { cartodb_id: 6, _cdb_feature_count: 1 },
+                                        { cartodb_id: 7, _cdb_feature_count: 1 },
+                                        { cartodb_id: 8, _cdb_feature_count: 1 },
+                                        { cartodb_id: 9, _cdb_feature_count: 1 }
                                     ];
+
+                                    // We check that the cells either added completely in the tile
```
Algunenano: > We check that the cells added

`We check that the cells are added either`

test/acceptance/aggregation.js
```diff
@@ -521,7 +524,7 @@ describe('aggregation', function () {
             });
 
             ['centroid', 'point-sample', 'point-grid'].forEach(placement => {
-                it('should provide all the requested columns in non-default aggregation ',
+                it('should provide all the requested columns in non-default aggregation: ' + placement,
```
jgoizueta: nice fix of the test title

test/acceptance/aggregation.js
Algunenano: It's slow in PROJ6. The technical reason is because st_buffer(geography) does 2 projection conversions, first to whatever SRS is most appropiate depending on the coordinates, then it does the buffer in that SRS and finally it goes back to 4326. In PROJ6 creating transformation is slower since it does more checks to be more precise, which isn't a big issue when the projection can be cached, but that's not the case for geographies (since the SRS changes from row to row).

lib/cartodb/models/aggregation/aggregation-query.js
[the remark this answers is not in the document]
dgaubert: I personally like the format used in aggregations: https://github.com/CartoDB/Windshaft-cartodb/blob/master/lib/cartodb/models/dataview/aggregation.js

But I left it to your wish as I don't have a strong opinion on that and we have other parts (histograms) with your indentation :)

lib/cartodb/models/aggregation/aggregation-query.js
Algunenano: I've looked around and it comes from OGC's Styled Layer Descriptor Implementation Specification:
```
The “standardized rendering pixel size” is defined to be 0.28mm × 0.28mm (millimeters).
Frequently, the true pixel size of the final rendering device is unknown in the web
environment, and 0.28mm is a common actual size for contemporary video displays.
```

PR #1114, opened 2019-07-22
Upgrade windshaft to version 5.5.1

dgaubert:


package-lock.json
```diff
@@ -166,7 +166,7 @@
     },
     "async": {
       "version": "1.5.2",
-      "resolved": "https://registry.npmjs.org/async/-/async-1.5.2.tgz",
+      "resolved": "http://registry.npmjs.org/async/-/async-1.5.2.tgz",
```
Algunenano: Please check your npm to avoid https -> http (we keep changing this with almost every update).

dgaubert: I've tried to force to use `https` but I guess It's an `npm` bug when I install a dependency that comes from github.

dgaubert: Not when installing a dependency with a GitHub's URL but upgrading it in this way: `npm remove windhsaft && npm install windshaft@5.5.1`

PR #1115, opened 2019-07-29
Update windshaft to version 5.6.0

dgaubert:


lib/cartodb/api/api-router.js
```diff
@@ -100,7 +100,9 @@ module.exports = class ApiRouter {
 
         const tileBackend = new windshaft.backend.Tile(rendererCache);
         const attributesBackend = new windshaft.backend.Attributes();
-        const previewBackend = new windshaft.backend.Preview(rendererCache);
+        const concurrency = serverOptions.renderer.mapnik.poolSize +
+                            serverOptions.renderer.mapnik.poolMaxWaitingClients;
+        const previewBackend = new windshaft.backend.Preview(rendererCache, { concurrency });
```
Algunenano: @dgaubert This looks like walking on thin ice. What would happen if the client has any other concurrent request? Wouldn't it make more sense to use just part of the total available pool?

dgaubert: > @dgaubert This looks like walking on thin ice. 

You are describing my life. :trollface: 

We need to put a limit somewhere. Large preview images can need hundreds of tiles, I tried to be as much performant as possible. If I see it can lead to an issue we can set a max concurrency to a more conservative value. Maybe just setting concurrency to `poolSize` value is enough.

dgaubert: I'll take a look at the metrics to see if the current value has sense or not

Algunenano: Did anything come out of this?

dgaubert: Metrics didn't tell anything relevant. As we hadn't any `render-capacity` error for `static` endpoints in the last 30 days. 

Up to you to open an issue to be tackled in RT. When we are in RT we can handle it properly.

PR #1135, opened 2019-10-29
Track mapconfig id / template id in logs

oleurud:


oleurud:
The idea is to add more info to follow all the mapconfig flow:
- know where the map config was created or instantiated
- know where a template was created and instantiated
- know the relation between template and mapconfig

Makes sense?

dgaubert:
> know where the map config was created or instantiated

dgaubert:
> know where the map config was created or instantiated
> know where a template was created and instantiated

Do you mean the Maps API instance? If so, we already have that info. See: `event_source_node ` field.

> know the relation between template and map-config

That's more interesting. 🤔

oleurud:
Let me explain better:
- know in which request the map config was created or instantiated (now we don't know that, it is important to know when the mapconfig is created)
- know in which request a template was created and instantiated
- know the relation between template id and mapconfig id

With all this info, we can know the lifecycle of a mapconfig / template. The idea is to get more info to investigate lost mapconfigs. 

I know I have added template id / mapconfig id in some places where we already have it (for example in the template update or delete where we get them from the URL) but I want to have a consistent place to have the info.

dgaubert:
Yeah, I know what you want to accomplish here. 😄 

Apart from the links that I shared above, which give you most of what you need to be able to add it in the log entry, you'll have to add a new header with the template id (`X-Template-Id`) when the template is created. Once you have it you need to use those headers to add them in the log entry. 

See: https://github.com/CartoDB/Windshaft-cartodb/blob/master/lib/api/template/admin-template-controller.js#L135-L144

lib/api/middlewares/logging-info.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+module.exports = function loggingInfo () {
+    // add to logs basic information about the request
+    // like: map config id
+    return function loggingInfoMiddleware (req, res, next) {
+        const info = {
+            mapConfigId: getMapconfigId(res.locals),
+            templateId: res.locals.templateId
```
dgaubert: You can get the Template Id from either `req.params` or `res.body`. No need to place it in `res.locals`.

lib/api/api-router.js
```diff
@@ -227,6 +228,8 @@ module.exports = class ApiRouter {
             this.templateRouter.route(apiRouter, route.template);
             this.mapRouter.route(apiRouter, route.map);
 
+            apiRouter.use(loggingInfoMiddleware());
```
dgaubert: Why only in this router?. It could be nice to have for every endpoint, couldn't be it?

lib/api/middlewares/logging-info.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+module.exports = function loggingInfo () {
+    // add to logs basic information about the request
+    // like: map config id
+    return function loggingInfoMiddleware (req, res, next) {
+        const info = {
+            mapConfigId: getMapconfigId(res.locals),
```
dgaubert: You can get it from `req.params` or `res.body`. No need to place it in `res.locals`.

lib/api/middlewares/logging-info.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+module.exports = function loggingInfo () {
```
dgaubert: what about `setTilerInfoHeader`?

lib/api/middlewares/logging-info.js
oleurud: Ok. I will use `res.locals` only in the creation

lib/api/api-router.js
oleurud: That was my idea. If I am not wrong, every request pass for it

lib/api/middlewares/logging-info.js
oleurud: Ok, makes sense

lib/api/api-router.js
dgaubert: Sorry I didn't notice you were using `apiRouter`, I thought it was bound to `templateRouter`

lib/api/middlewares/tiler-info-header.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+const { templateName } = require('../../backends/template-maps');
+
+module.exports = function setTilerInfoHeader() {
+    // add basic information about the request to logs
+    return function setTilerInfoHeaderMiddleware(req, res, next) {
+        const info = {
+            mapConfigId: getMapconfigId(res.locals),
```
dgaubert: Same as template, get it from `req.params` or `res.body.layergroupid`. Also avoid:
```
{
  mapConfigId: undefined
  ...
}

File: lib/api/template/tile-template-controller.js [hunk 6]
```diff
@@ -82,6 +82,7 @@ function getTile ({ tileBackend, label }) {
 
             res.statusCode = 200;
             res.body = tile;
+            res.locals.mapConfig = mapConfigProvider.mapConfig;
```
dgaubert: we don't need it, do we?

lib/api/middlewares/named-map-provider.js
```diff
@@ -27,6 +27,7 @@ module.exports = function getNamedMapProvider ({ namedMapProviderCache, label, f
             }
 
             res.locals.mapConfigProvider = namedMapProvider;
+            res.locals.templateId = template_id;
```
dgaubert: We don't need it, do we?

oleurud: I forgot it

lib/api/middlewares/tiler-info-header.js
oleurud: > But we can not have the info in tile requests for example.

I did not realize that we could get it from `req.params`.

It seems very complicated in my opinion, but I going to try that way.

oleurud: I am not going to use `req.params.token` because it is already clean in `res.locals.token`

To use `res.body.layergroupid` I need to do:
```
    if (body.layergroupid) {
        const layergroupid = body.layergroupid;
        if (layergroupid.includes(':')) {
            return layergroupid.split(':')[0];
        }

        return layergroupid;
    }
```
Is it not simpler to get the id from mapconfig object?

oleurud: For the named cases, I will need to do more things. Starting for knowing when we are in that case or changing the code to use the `setLayergroupIdHeaderMiddleware` in more places...

lib/api/template/tile-template-controller.js
oleurud: As I explained in the other comment, it seems more complicated to use the logic from `setLayergroupIdHeaderMiddleware` instead of saving mapconfig in locals (or simply the id if you see any memory problem)

lib/api/middlewares/tiler-info-header.js
dgaubert: The timestamp is part of the layergroup token, we should not skip it as it's relevant information for debugging. In Kibana you can filter using Regex.

lib/api/template/tile-template-controller.js
dgaubert: you already have the layergroup token in multiple places:
  - The raw token: req.params.token.
  - The layergroup and the timestamp: https://github.com/CartoDB/Windshaft-cartodb/blob/master/lib/api/middlewares/layergroup-token.js#L13-L14

dgaubert: Sorry, I was mixing controllers we are in template.

dgaubert: You should use `mapConfigProvider.getMapConfig()` method instead.

dgaubert: In fact, why not move this to `setTilerInfoHeader` middleware as `mapConfigProvider` is already set in `res.locals` which is available in that middleware?.

lib/api/middlewares/tiler-info-header.js
dgaubert: > About ´undefined´, kibana filters these values.

Let's keep the log entry as much cleaner as possible. It adds some bytes per log entry and wastes disc space

oleurud: Yes, it is. And I agree it could be interesting, but what this PR tries to achieve is to start logging the mapconfig id in all requests where it exists. If we would want to log the layergroup id, we should ensure in every request and it is not the objective

lib/api/template/tile-template-controller.js
oleurud: Ok, I am going to extract the mapconfig id from `mapConfigProvider` in `setTilerInfoHeader`

> You should use mapConfigProvider.getMapConfig() method instead.

As `setTilerInfoHeader` happens after all the controller middlewares if everything goes fine, `mapConfigProvider` should already have the `mapConfig` property set and we can avoid doing a new redis request.

oleurud: I realized for the same reason, I can use `getMapConfig()` (because the method returns it directly if it already have it without doing a new redis request) so I am going to use it (as we are doing in the rest of the code)

dgaubert: It's already cached after the first fetch, see:
- https://github.com/CartoDB/Windshaft-cartodb/blob/master/lib/models/mapconfig/provider/map-store-provider.js#L37-L39
- https://github.com/CartoDB/Windshaft-cartodb/blob/master/lib/models/mapconfig/provider/named-map-provider.js#L58-L60

Please, use the public interface methods. The member `.mapConfig` should be private...

PR #1144, opened 2019-11-26
Add layergroupid to tiler-stats 

dgaubert:
Add `layergroupid` to tiler-stats  to be able to correlate different requests into the same workflow


dgaubert:
Closing, we need to analyze a complete approach to correlate metrics

PR #1153, opened 2020-02-21
Update free limits

gonzaloriestra:
Related to: https://github.com/CartoDB/product/issues/530

The new Free price plan will have the same limits as the Individual plan, so we have to update the developer center.


gonzaloriestra:
We are finally leaving the original limits.

PR #1154, opened 2020-02-25
Add pubsub metrics

esloho:
Closes [ch58706] https://app.clubhouse.io/cartoteam/story/58706/add-pubsub-middleware-to-maps-api

This PR adds a new middleware for sending metrics as pubsub events. This new middleware should be asynchronous and not wait for pubsub to finish.

**Acceptance:**

- Regular behavior should not be altered (navigate and perform some queries on staging to check everything goes well). No events should be generated since no request would add the headers now.
- Run some curl/postman requests with the new headers and check new events are created in the staging project in google cloud (ping me if you don't have access).

### DIFF
--- a/lib/api/middlewares/user.js
+++ b/lib/api/middlewares/user.js
@@ -2,12 +2,30 @@
 
 const CdbRequest = require('../../models/cdb-request');
 
-module.exports = function user () {
+module.exports = function user (metadataBackend) {
     const cdbRequest = new CdbRequest();
 
     return function userMiddleware (req, res, next) {
-        res.locals.user = cdbRequest.userByReq(req);
+        res.locals.user = getUserNameFromRequest(req, cdbRequest);
 
-        next();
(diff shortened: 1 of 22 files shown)


lib/api/middlewares/user.js
```diff
@@ -2,12 +2,30 @@
...
-module.exports = function user () {
+module.exports = function user (metadataBackend) {
     const cdbRequest = new CdbRequest();
 
     return function userMiddleware (req, res, next) {
-        res.locals.user = cdbRequest.userByReq(req);
+        res.locals.user = getUserNameFromRequest(req, cdbRequest);
 
-        next();
+        getUserId(metadataBackend, res.locals.user, function (userId) {
```
oleurud: A new redis request for each map request... I'm not sure it's worth it

esloho: We need to extract the user_id for the pubsub events, since we're not using usernames for metrics :/

PR #1155, opened 2020-03-04
Add response time to pubsub events

esloho:
Closes [ch61185] https://app.clubhouse.io/cartoteam/story/61185/add-response-time-to-apis-events

### DIFF
--- a/lib/api/middlewares/pubsub-metrics.js
+++ b/lib/api/middlewares/pubsub-metrics.js
@@ -41,6 +41,12 @@ function getEventData (req, res) {
         attributes.event_group_id = eventGroupId;
     }
 
+    const responseTime = getResponseTime(res);
+
+    if (responseTime) {
+        attributes.response_time = responseTime.toString();
+    }
+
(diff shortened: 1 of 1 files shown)


lib/api/middlewares/pubsub-metrics.js
```diff
@@ -52,4 +58,17 @@ function normalizedField (field) {
     return field.toString().trim().substr(0, MAX_LENGTH);
 }
 
+function getResponseTime (res) {
+    const profiler = res.get('X-SQLAPI-Profiler');
```
oleurud: I would prefer to get it from `req.profiler.toJSONString()`, is it possible?

If not, the header is `X-Tiler-Profiler`

esloho: I first tried getting it directly from the profiler but then the time would differ slightly, so I thought was better to stick with the already calculated version to avoid discrepancies.

Sorry for the bad header :woman_facepalming:

PR #1161, opened 2020-04-09
Fixes 1160: Prevent using cast column as part of __ctx_query

### LINKED ISSUE #1160: Error parsing columns in numeric histograms
## Context

When I add a time-series widget based on a date column Windshaft returns an error:
```json
{
  "errors": [
    "schema \"__ctx_query\" does not exist"
  ],
  "errors_with_context": [
    {
      "type": "unknown",
      "message": "schema \"__ctx_query\" does not exist"
    }
  ]
}
```
After debugging Windhaft, it seems it tries to run a query which has the following:

```sql
WITH  
 __cdb_basics AS ( 
    ...
FROM 
( 
    SELECT * FROM (SELECT * FROM (SELECT * FROM multi_location) _cdb_bbox_filter 
WHERE ST_Intersects(the_geom_webmercator, ST_Transform(ST_MakeEnvelope(-124.08690690994264,12.740371611587053,-47.09471940994263,50.77352319541077, 4326), 3857))) __ctx_query, __cdb_basics  
              WHERE __ctx_query.date_part( 'epoch ', kickoff) >= 1577836800 
                AND __ctx_query.date_part( 'epoch ', kickoff) <= 1701388800 
             
) __cdb_filtered_source_query, __cdb_basics 
GROUP BY 10, __cdb_basics.__cdb_bins_number 
ORDER BY 10;
```

Which is incorrect because it executes the function `date_part` as part of `__ctx_query` which is not an schema but an alias for a subquery.

This is because in the function [_buildQuery](https://github.com/CartoDB/Windshaft-cartodb/blob/69afee61e0165fd976dc9becb9cb0dff10f5d67d/lib/models/dataview/histograms/numeric-histogram.js#L56) of the class NumericHistogram cast the column if it is date type and then it uses it in [_buildQueryTpl](https://github.com/CartoDB/Windshaft-cartodb/blob/69afee61e0165fd976dc9becb9cb0dff10f5d67d/lib/models/dataview/histograms/numeric-histogram.js#L90) as part of the alias `__ctx_query`.

manmorjim:
Prevent using cast column when it is date type as part of the alias `__ctx_query` for numeric histograms by keeping the original column.

Fixes #1160

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -21,6 +21,9 @@ Announcements:
 - Remove `bootstrapFonts` at process startup (now done in `windshaft@6.0.0`)
 - Stop checking the installed version of some dependencies while testing
 
+Bug Fixes:
+- Parsing date column in numeric histograms (#1160)
+
 ## 8.1.1
 Released 2020-02-17
 
--- a/lib/models/dataview/histograms/numeric-histogram.js
+++ b/lib/models/dataview/histograms/numeric-histogram.js
@@ -52,10 +52,18 @@ Numeric histogram:
 */
 module.exports = class NumericHistogram extends BaseHistogram {
     _buildQuery (psql, override, callback) {
+        let column = this.column;
+        let query = this.query;
+        // for date type we have to cast the column using an alias
+        // and using that alias to prevent multiple calls to the cast
+        if (this._columnType === 'date') {
+            query = `(SELECT *, ${utils.columnCastTpl({ column })} as __cdb_cast_date FROM (${this.query}) __cdb_original_query)`;
+            column = '__cdb_cast_date';
+        }
         const histogramSql = this._buildQueryTpl({
-            column: this._columnType === 'date' ? utils.columnCastTpl({ column: this.column }) : this.column,
+            column,
             isFloatColumn: this._columnType === 'float',
-            query: this.query,
+            query,
             start: this._getBinStart(override),
             end: this._getBinEnd(override),
             bins: this._getBinsCount(override),
--- a/test/acceptance/dataviews/histogram-test.js
+++ b/test/acceptance/dataviews/histogram-test.js
@@ -325,6 +325,15 @@ describe('histogram-dataview for date column type', function () {
                     column: 'd',
                     aggregation: 'minute'
                 }
+            },
+            date_histogram_no_agg: {
+                options: {
+                    column: 'd'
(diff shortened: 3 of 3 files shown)


Algunenano:
Why not add a new column to the query with the casted date (__cdb_casted_date) and use that for the aggregation?

Also, we should add a test for this.

manmorjim:
> Why not add a new column to the query with the casted date (__cdb_casted_date) and use that for the aggregation?

:thinking: We have no control over the `query` of the `ctx` object in this class because it is given as parameter from an external call. However it could be addressed by modifying the line [139](https://github.com/CartoDB/Windshaft-cartodb/blob/fix_parsing_columns_histograms_1160/lib/models/dataview/histograms/numeric-histogram.js#L139) with:

```js
`
FROM
(
    SELECT * FROM (SELECT *, ${ctx.column} as __cdb_cast_date FROM (${ctx.query})) __ctx_query${extraTables} ${extraFilter}
)
`
```

And then in [extraFilter](https://github.com/CartoDB/Windshaft-cartodb/blob/fix_parsing_columns_histograms_1160/lib/models/dataview/histograms/numeric-histogram.js#L99):

```js
extraFilter = `
  WHERE __ctx_query.__cdb_cast_date >= ${ctx.start}
    AND __ctx_query.__cdb_cast_date <= ${ctx.end}
`;
```

What do you think about this approach?


> Also, we should add a test for this.

You're right. I'll mark this PR as draft until I add a test :+1:

Algunenano:
> thinking We have no control over the query of the ctx object in this class because it is given as parameter from an external call. However it could be addressed by modifying the line 139 with:

I'd modify https://github.com/CartoDB/Windshaft-cartodb/blob/fix_parsing_columns_histograms_1160/lib/models/dataview/histograms/numeric-histogram.js#L61 if we need it and set https://github.com/CartoDB/Windshaft-cartodb/blob/bb9d8f1385c817e925f35da43486a171c932c6c5/lib/models/dataview/histograms/numeric-histogram.js#L57 (column) to the new created column, that way there is no need to modify the query and we avoid the constant calls to the cast, right?

manmorjim:
Thank you both! my first contribution to windshaft \o/ :tada: :rocket:

dgaubert:
:shipit:

PR #1172, opened 2020-06-03
Stop using profiler wrongly

dgaubert:
Stop using profiler wrongly. Now it only saves custom events from backends (tile, map, attributes, etc..) and calculates the response time. Besides, removed tags to know whether overviews are being used.


dgaubert:
I need to merge it now to continue working on this. Please @oleurud leave any comment when you are able to and I'll apply changes in the next PR.

PR #1173, opened 2020-06-04
Logger improvements

dgaubert:
- Centralize common headers, this will help us to later move metrics out of the process.
- Be able to reduce the footprint in the final log file depending on the environment


lib/utils/logger.js
```diff
@@ -1,12 +1,17 @@
...
-const { req: requestSerializer, res: responseSerializer, err: errorSerializer } = pino.stdSerializers;
+const { req: requestSerializer, res: responseSerializer, err, wrapErrorSerializer } = pino.stdSerializers;
+const DEV_ENVS = ['test', 'development'];
 
 module.exports = class Logger {
     constructor () {
         const { LOG_LEVEL, NODE_ENV } = process.env;
         const logLevelFromNodeEnv = NODE_ENV === 'test' ? 'fatal' : 'info';
+        const errorSerializer = DEV_ENVS.includes(NODE_ENV) ? err : wrapErrorSerializer(err => {
+            err.stack = err.stack.split('\n').slice(0, 3).join('\n');
```
oleurud, line 12: Why only the last three stack lines? Not sure, having the full trace can help to get some insights when an error rises

lib/utils/common-headers.js
```diff
@@ -0,0 +1,127 @@
+'use strict';
+
+module.exports = function setCommonHeaders (req, res, callback) {
```
oleurud, line 3: This is a really good addition :) 

Just a proposal: we could promisify it
- promisify getStatTag function
- promisify this one

With that, we can avoid receiving and calling the callback from here. I think the callers would be cleaner that way

lib/utils/logger.js
dgaubert: Not the last three but the first three :). I'm concerned about the amount of info that ELK will receive from the log file, I was trying to reduce the total footprint, for instance:

- Full stack trace (~1900 bytes):
```json
{"level":40,"time":1591295177114,"id":"c8c0a798-cd6c-4c5a-b874-086d1b11dbda","error":[{"type":"Error","message":"Limit for number of rows is not defined","stack":"Error: Limit for number of rows is not defined\n    at Object.limitNumberOfRows (/vagrant/repos/Windshaft-cartodb/node_modules/camshaft/lib/node/limits.js:11:21)\n    at BaseNode.Node.checkLimits (/vagrant/repos/Windshaft-cartodb/node_modules/camshaft/lib/node/node.js:537:12)\n    at /vagrant/repos/Windshaft-cartodb/node_modules/camshaft/lib/limits/checker.js:12:18\n    at /vagrant/repos/Windshaft-cartodb/node_modules/async/lib/async.js:181:20\n    at iterate (/vagrant/repos/Windshaft-cartodb/node_modules/async/lib/async.js:262:13)\n    at Object.async.forEachOfSeries.async.eachOfSeries (/vagrant/repos/Windshaft-cartodb/node_modules/async/lib/async.js:281:9)\n    at Object.async.forEachSeries.async.eachSeries (/vagrant/repos/Windshaft-cartodb/node_modules/async/lib/async.js:214:22)\n    at module.exports (/vagrant/repos/Windshaft-cartodb/node_modules/camshaft/lib/limits/checker.js:9:11)\n    at analysis$checkLimits (/vagrant/repos/Windshaft-cartodb/node_modules/camshaft/lib/analysis.js:91:17)\n    at fn (/vagrant/repos/Windshaft-cartodb/node_modules/async/lib/async.js:746:34)\n    at /vagrant/repos/Windshaft-cartodb/node_modules/async/lib/async.js:1213:16\n    at /vagrant/repos/Windshaft-cartodb/node_modules/async/lib/async.js:166:37\n    at /vagrant/repos/Windshaft-cartodb/node_modules/async/lib/async.js:706:43\n    at /vagrant/repos/Windshaft-cartodb/node_modules/async/lib/async.js:167:37\n    at Immediate._onImmediate (/vagrant/repos/Windshaft-cartodb/node_modules/async/lib/async.js:1206:34)\n    at processImmediate (internal/timers.js:456:21)","node":{"id":"a23b8472f12aa0c4caea27c8c7a39b49d85e1b77","type":"source","query":"SELECT * FROM populated_places_simple_reduced"}}]}
```
- Reduced stack trace (~500 bytes):
```json
{"level":40,"time":1591293920140,"id":"0d13fccd-3cc5-4442-ab14-82c20519077f","error":[{"type":"Error","message":"Limit for number of rows is not defined","stack":"Error: Limit for number of rows is not defined\n    at Object.limitNumberOfRows (/vagrant/repos/camshaft/lib/node/limits.js:12:21)\n    at BaseNode.Node.checkLimits (/vagrant/repos/camshaft/lib/node/node.js:537:12)","node":{"id":"a23b8472f12aa0c4caea27c8c7a39b49d85e1b77","type":"source","query":"SELECT * FROM populated_places_simple_reduced"}}]}
```

Looking at the two stack traces:

- Full stack trace:
```
```

lib/utils/common-headers.js
dgaubert: As `setCommonHeaders` is called from a middleware `function (req, res, next) {...}`, for the moment, we can't use `async/await` and I have to use `.then` and/or `.catch`. The caller wouldn't be as clean as we want:

```js
setCommonHeaders(req, res)
    .then(() => {
        const errorResponseBody = {
            errors: errors.map(errorMessage),
            errors_with_context: errors.map(errorMessageWithContext)
        };
        [ ... ]
        return next();
    });
```

I'll wait for express 5 to rewrite middlewares and collaborators that use callbacks to promises to be able to use `async`/`await`.

lib/utils/logger.js
Algunenano: Why make the trade-off? We shouldn't have any stack trace right in the logs unless it's a programming error, right? And in those cases more information is useful.

dgaubert: NOTE: I've copied those log entries from tests that I ran yesterday. The second one was using `npm link camshaft` that's why the paths are different. :)

dgaubert: > Why make the trade-off? We shouldn't have any stack trace right in the logs unless it's a programming error, right? And in those cases more information is useful.

Yeah, those errors aren't programming errors, but still relevant to know from where that error is coming from as many errors will come from different modules (camshaft, windshaft, cartonik, grainstore, etc..) and error message might not be accurate enough.

I've thought in that approach, removing stack traces completely as we  (@oleurud, you and me) are able to figure out where errors are coming from w/o looking at stack traces. However, I remembered the main goal of revamping the logging system: anyone at CARTO should have enough info to debug customer's issues, such as CS, Support, and RT.

I think the stack trace: `Error: Limit for number of rows is not defined at Object.limitNumberOfRows (/vagrant/repos/camshaft/lib/node/limits.js:12:21)` will point anyone at CARTO to the right direction when a user is claiming because his map with analysis is taking too long to load. Don't you?

dgaubert: To recap:
- @oleurud: full stack trace.
- @dgaubert: reduced stack trace.
- @Algunenano: no stack trace.

If we were four developers I'm pretty sure we'd have four different approaches. 😂

oleurud: I should say I have never look at the trace in production when an error happens. But we had looked at the production logs from time to time (ok, just one time) to get a list of errors happening in production that we are not aware of.

I think they're valuable to all of us since we don't have a platform monitoring errors, so this is the only way to know errors that are not reported by clients. And we all know, there are a few of those :)

lib/utils/common-headers.js
oleurud: I am thinking in somthing like:

```
module.exports = function sendResponse () {
  return async function sendResponseMiddleware (req, res, next) {
    ...
    await setCommonHeaders(req, res).catch(next)
    ...
  }
}
```

oleurud: Mixing await/async and callbacks are not the best, but it makes sense for me

oleurud: ouch, forget it, it is not possible

PR #1175, opened 2020-06-08
Tags Middleware

dgaubert:
Be able to tag requests with labels as easier way to provide business metrics


lib/api/middlewares/tag.js
```diff
@@ -0,0 +1,15 @@
...
+
+module.exports = function tag ({ tags }) {
+    if (!Array.isArray(tags) || !tags.every((tag) => typeof tag === 'string')) {
+        throw new Error('Required "tags" option must be a valid Array: [string, string, ...]');
+    }
+
+    return function tagMiddleware (req, res, next) {
+        const { logger } = res.locals;
+        res.locals.tags = tags;
+        res.on('finish', () => logger.info({ tags: res.locals.tags }));
```
oleurud, line 11: Why don't you log it immediately?

dgaubert: Good question. For the same reason that we are assigning them to a `res.locals`: because we want to have the ability to add more labels in the next middlewares. For instance:

```js
function getTileMiddleware(req, res, next) {
  const { format } = req.params;
  console.log(res.locals.tags); // [ 'tile' ]
  res.locals.tags.push(format);
}

// then will log: [ 'tile', 'png' ]
```

PR #1178, opened 2020-06-09
Layergroup Id should have cache buster defined always

dgaubert:
Set a cache buster equal to `0` when there are no affected tables in the MapConfig. Thus `layergroupid` would have the same structure always:
  - `${map_id}:${cache_buster}` for anonymous map
  - `${user}@${template_hash}@${map_id}:${cache_buster}` for named map

Needed to facilitate parsings in other places in our platform (metrics, logs, etc..)

cc @ilbambino

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 9.0.1
+Released 2020-mm-dd
+
+Bug Fixes:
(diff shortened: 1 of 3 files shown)


Algunenano:
A test so we don't break this in the future would be great!

dgaubert:
👮

dgaubert:
> A test so we don't break this in the future would be great!

Done.

PR #1180, opened 2020-06-24
changes in tests to work in docker-dev-env

juanrmn:


dgaubert:
Yeah, nice work @juanrmn. it makes sense. 

Even I'd go further and I'd require both Redis and Postgres up and running in their respective ports.

Thanks for sharing it.

PR #1181, opened 2020-06-29
Fix: TypeError: Cannot read property 'level' of undefined

dgaubert:
Feature: dump unfinished log into a file while exiting

cc/ @jvillarf

### DIFF
--- a/metro/log-collector.js
+++ b/metro/log-collector.js
@@ -1,10 +1,13 @@
 'use strict'
 
+const fs = require('fs');
 const split = require('split2');
 const assingDeep = require('assign-deep');
 const { Transform } = require('stream');
 const DEV_ENVS = ['test', 'development'];
-const logs = new Map();
+const dumpPath = `${__dirname}/dump.json`;
+
+let logs;
 
 const LEVELS = {
     10: 'trace',
@@ -16,7 +19,7 @@ const LEVELS = {
 }
 
 module.exports = function logCollector () {
-    return new Transform({
+    const stream = new Transform({
         transform (chunk, enc, callback) {
             let entry;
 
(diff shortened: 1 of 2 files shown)


metro/log-collector.js
```diff
@@ -34,25 +37,25 @@ module.exports = function logCollector () {
...
+            const { id } = entry;
 
             if (id === undefined) {
                 entry.level = LEVELS[entry.level];
                 this.push(`${JSON.stringify(entry)}\n`);
                 return callback();
             }
 
-            if (end === true) {
-                const accEntry = logs.get(id);
```
oleurud, line 46: We missed it

metro/log-collector.js
```diff
@@ -70,7 +73,32 @@ module.exports = function logCollector () {
 
             callback();
         }
-    })
+    });
+
+    stream.on('pipe', () => {
```
oleurud, line 78: Not very sure about the duality in memory, in file. I would prefer to use one or the other. Let's see how it works.

But my real question is: if we are going to save the current request log to a file when app dies, why do we want to restore it when the app is up again? For instance, if the app dies, the current request with id = '1234' will die too, so I guess `logs.has('1234')` is not going to be true again anymore (or even worst if yes, we will merge 2 different requests in the same request log)

dgaubert: In production, we are using a socket file FIFO: https://www.freedesktop.org/software/systemd/man/systemd.socket.html#ListenFIFO=

So when it dies the non processed logs are accumulated in the socket file, when it restarts it will consume the non processed logs from the socket file.

oleurud: I thought metro was getting the intermediate logs from the current request (from the main express app), but metro is receiving them from stdin. 

So, ok, my concern is clear then.

jvillarf:
:clap:

PR #1190, opened 2020-09-09
Fix logger error serializer when the exception stack is not set

thedae:
Related RT story: https://app.clubhouse.io/cartoteam/story/101625/node-windshaft-exiting-because-of-typeerror-cannot-read-property-split-exception


thedae:
Not ready yet

dgaubert (approved):
LGTM.

We should avoid logging non-error instances when using the error serializer (`logger.log({ error: nonErrorInstance })`). I'm ok this check FTM and fix the issue in production.

🚀

PR #1196, opened 2021-01-14
undefined might come as a string

ilbambino:
so a hack to check and cast it to a real undefined


alberhander:
Tested it, this fixes the wrong windshaft urls in dev-env.

alberhander (approved):

PR #1197, opened 2021-01-26
Replace analysis static config per ENV vars

thedae:
The analysis host now uses env vars instead of static config. This fixes both builder analysis and batch API requests in local environment


ilbambino (approved):

alberhander (approved):
LGTM!

PR #1199, opened 2021-02-11
Adapt service for the new development environment (no 3rd party tools required)

dgaubert:
In order to run tests in the new development environment:
- stop using pg tools (`psql`, `pg_dump`, ...) and redis-cli to setup fixtures
- clean tests leftovers
- adapt SQL scripts to be run with `node-postgres`
- adapt Redis set up to be run with `node-redis` 
- fix tests where they create keys in Redis that weren't cleaned

Note: I guess we can stop using dedicated images for testing purposes


test/support/test-helper.js
```diff
@@ -69,77 +69,48 @@ function checkSurrogateKey (res, expectedKey) {
...
+    for (const db of Object.keys(expectedKeys)) {
+        await new Promise((resolve, reject) => client.select(db, (err) => err ? reject(err) : resolve()));
+        const keys = await new Promise((resolve, reject) => client.keys('*', (err, keys) => err ? reject(err) : resolve(keys)));
+        foundKeys[db].push(...keys);
+    }
 
-        if (Object.keys(databasesTasks).length === 0) {
-            assert.strictEqual(keysFound.length, 0, 'Unexpected keys found in redis: ' + keysFound.join(', '));
-            done();
+    const unexpectedKeys = { 0: [], 5: [] };
```
oleurud, line 98: should we add here 8 and 3?

oleurud: Answering myself, I think the rate limit keys (db 8) are well removed in tests, but maybe is a good thing, in any case, to check them here

oleurud (approved):
I see it fine. I added 2 "just in case questions". 

I have not checked the SQL files, since tests are working I guess fixtures are fine (and that the change has not been too painful for you)

test/index.js
```diff
@@ -56,87 +71,131 @@ async function populateDatabase () {
...
+        '_vovw_1_test_special_float_values_table_overviews',
+        'test_table_localhost_regular1',
+        'analysis_banks',
+        'analysis_rent_listings',
+        'test_table_100',
+        'test_table_200k',
+        'populated_places_simple_reduced',
+        'populated_places_simple_reduced_private'
+    ];
+    await query({ db: TEST_DB, sql: `VACUUM ANALYZE ${tables.join(', ')}` });
```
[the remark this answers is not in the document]
dgaubert: To stabilize tests, it's legacy. Not sure if we still need it.

test/support/test-helper.js
dgaubert: Yeah, I'm going to check them, the main goal is to leave Redis as it was before running the tests.

test/index.js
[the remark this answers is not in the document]
oleurud: Ahh, I didn't remember that I did, so I didn't say anything.

test/support/test-helper.js
dgaubert: the DB 3, do you know for what it is being used?

oleurud: Old "oauth" stuff https://github.com/CartoDB/node-cartodb-redis/blob/master/lib/carto_metadata.js#L48

Probably it is not in use, at least I know nothing about it.

oleurud: It seems to be used in SQL API https://github.com/CartoDB/CartoDB-SQL-API/blob/master/lib/auth/oauth.js, but not here

ilbambino (approved):
I like the changes, 👍 💯

PR #1222, opened 2021-10-29
Add notification when user is reaching named maps limit

moicalcob:
There are several users that have reached their named map limits, and we haven't been notified about that. I've added a check when creating a new named map to ensure that the user is not reaching the named maps limit.

If the user is near the limit, the API will send a response with a notification message.


lib/backends/template-maps.js
```diff
@@ -195,6 +195,32 @@ TemplateMaps.prototype._checkUserTemplatesLimit = function (userTemplatesKey, ow
...
+    if (!limit) {
+        return callback();
+    }
+
+    this._redisCmd('HLEN', [userTemplatesKey], (err, numberOfTemplates) => {
+        if (err) {
+            return callback(err);
+        }
+
+        if (numberOfTemplates >= limit - 100) {
```
Shylpx, line 221: I would recommend storing the threshold value (100) in a constant.

Shylpx (approved):
:rocket:

PR #1230, opened 2022-08-03
Avoid massive notifications about named maps limit

moicalcob:
Last months we introduced a new notification when a user was reaching the named map limits.
With this change we're going to start saving in Redis the last date that we notified for a user about this limit to avoid sending massive notifications through email. If a notification has been sent in the last month, we'll omit the notification!


Shylpx (approved):
Nice job! :rocket:

